PR #1, opened 2023-06-10
Implement the driver

uuf6429:
Remaining Tasks
- ~~Update readme~~ #4
- [ ] Remove/replace syn.js if possible (https://github.com/uuf6429/webdriver-classic-driver/pull/2, #5)
- [x] More tests
  - [ ] Checking codecov coverage should help
- [x] Remove extra public api (done: d385eed7f7ec955f3706d765003a5907beb85509)
  - [ ] Discuss adding it back officially
- [x] Investigate random failures ([example1](https://github.com/uuf6429/webdriver-classic-driver/actions/runs/5281495675/jobs/9555153221), [example2](https://github.com/uuf6429/webdriver-classic-driver/actions/runs/5281653633/jobs/9559640651))
     seems they're always about seleium4+chrome?
     Couldn't reproduce them, and last two runs were failure-free - maybe problem got solved somehow?


phpunit.xml.dist
```diff
@@ -18,15 +19,18 @@
...
+
         <!-- where driver will connect to -->
         <server name="DRIVER_URL" value="http://localhost:4444/wd/hub"/>
+
         <!-- where DocumentRoot of 'Test Machine' is mounted to on 'Driver Machine' (only if these are 2 different machines) -->
-        <!--server name="DRIVER_MACHINE_BASE_PATH" value="" /-->
-        <!--server name="TEST_MACHINE_BASE_PATH" value="" /-->
+        <server name="DRIVER_MACHINE_BASE_PATH" value="/fixtures/"/>
+
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="disabled" />
```
uuf6429, line 33: @stoff I wrote my thoughts [on this already](https://github.com/minkphp/MinkSelenium2Driver/issues/366#issuecomment-1585666969), but let's maybe let's start a proper discussion here?

Maybe I'm missing something, but isn't *display errors* + *enabling all errors* + *disallowing output from tests* good enough (and maybe also the expected flow)?

tests/bootstrap.php
```diff
@@ -0,0 +1,18 @@
...
+use Symfony\Component\Process\Process;
+
+$minkTestServer = new Process([
+    PHP_BINARY,
+    '-S',
+    '0.0.0.0:8002',
+    '-t',
+    __DIR__ . '/../vendor/mink/driver-testsuite/web-fixtures'
+]);
+$minkTestServer->start();
```
uuf6429, line 14: I find this more convenient than adding commands to the workflows and more steps in the documentation.

Any thoughts of possible downsides? Should it be documented somewhere?

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
+    /**
+     * {@inheritdoc}
+     * @throws DriverException
+     */
+    public function start(): void
+    {
+        try {
+            $this->webDriver = RemoteWebDriver::create($this->webDriverHost, $this->desiredCapabilities);
+            $this->applyTimeouts();
+            $this->initialWindowName = $this->getWindowName();
```
uuf6429, line 94: "Borrowed" this concept from [oleg-andreyev/MinkPhpWebDriver](https://github.com/oleg-andreyev/MinkPhpWebDriver) instead of deprecating the usage. cc @oleg-andreyev

phpunit.xml.dist
stof: I don't think what you describe will be enough.
Also, symfony/phpunit-bridge enables the DebugClassLoader when available, which reports additional deprecations based on class definitions.
I would prefer if this repo was using the listener of the bridge, as done in other drivers (note that we are not using the simple-phpunit that manages a separate phpunit installation. We are still installing phpunit as a composer dependency).

uuf6429: ok then , I'll add it back 👍

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
 
-/*
+/**
  * This file is part of the Behat\Mink.
  * (c) Konstantin Kudryashov <ever.zet@gmail.com>
  *
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
+ *
+ * @noinspection PhpLanguageLevelInspection
```
stof, line 10: why this ? Let your phpstorm configure its language level based on the composer.json and it should be enough.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
+        $this->setDesiredCapabilities($this->initCapabilities($desiredCapabilities ?? []));
+        $this->webDriverHost = $webDriverHost ?? 'http://localhost:4444/wd/hub';
+        $this->xpathEscaper = $xpathEscaper ?? new Escaper();
+    }
+
+    // <editor-fold desc="Implementation">
+
+    /**
+     * {@inheritdoc}
+     * @throws DriverException
```
stof, line 87: this is useless. The interface already defines this `@throws`

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
+            throw new DriverException("Could not open connection: {$e->getMessage()}", 0, $e);
+        }
+    }
+
+    /**
+     * Sets the timeouts to apply to the webdriver session
+     *
+     * @param array $timeouts The session timeout settings: Array of {script, implicit, page} => time in milliseconds
+     * @throws DriverException
+     * @api
```
stof, line 106: `@api` should be removed. Any non-private method is covered by BC unless tagged as `@internal`.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
+    {
+        $this->timeouts = $timeouts;
+
+        if ($this->isStarted()) {
+            $this->applyTimeouts();
+        }
+    }
+
+    /**
+     * {@inheritdoc}
```
stof, line 118: Any phpdoc containing only `{@inheritdoc}` is totally useless.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
+        foreach ($nodes as $i => $node) {
+            $elements[] = sprintf('(%s)[%d]', $xpath, $i + 1);
+        }
+
+        return $elements;
+    }
+
+    /**
+     * {@inheritdoc}
+     * @throws NoSuchElementException
```
stof, line 335: this is wrong. the interface says that methods should throw a DriverException when the operation cannot be done.

.github/workflows/ci.yml
```diff
@@ -23,39 +23,61 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
+    # We cannot trigger events in old chrome, so we allow the run to fail
+    continue-on-error: ${{ matrix.browser == 'chrome' && matrix.selenium == '2.53.1' }}
```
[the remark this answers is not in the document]
uuf6429: Right, I think I'll go with the test-skipping approach. 👍

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
+    private string $webDriverHost;
+
+    private ?string $initialWindowName = null;
+
+    /**
+     * @throws DriverException
+     */
+    public function __construct(
+        string  $browserName = null,
+        array   $desiredCapabilities = null,
```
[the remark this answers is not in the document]
uuf6429: I don't like it either, to be honest.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
+    {
+        return $this->desiredCapabilities->toArray();
+    }
+
+    /**
+     * Globally press a key i.e. not typing into an element.
+     *
+     * @api
+     */
+    public function globalKeyPress($char, $modifier = null): void
```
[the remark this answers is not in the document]
uuf6429: There are quite a few of those .. yes, let's discuss. The way forward - I'll remove them for now.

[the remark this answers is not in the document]
stof: If your grouping is right, there are 2 of them only: `globalKeyPress` and `dragBy` (the other ones are exposing the configuration of the driver, which might be discussed but is not something meant to be in the abstraction layer)

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
+     * @param string $script the script to execute
+     * @param Boolean $sync whether to run the script synchronously (default is TRUE)
+     * @return mixed
+     * @example $this->executeJsOnXpath($xpath, 'return argument[0].childNodes.length');
+     */
+    private function executeJsOnElement(
+        RemoteWebElement $element,
+        #[Language('JavaScript')]
+        string           $script,
+        bool             $sync = true
```
stof, line 1110: according to the code coverage, this method is never used in an async way. So it should be removed

tests/Custom/TimeoutTest.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $this->expectException(DriverException::class);
+        $this->getSession()->start();
+
+        $this->getSession()->getDriver()->setTimeouts(['invalid' => 0]);
+    }
+
+    public function testShortTimeoutDoesNotWaitForElementToAppear(): void
+    {
+        $this->getSession()->getDriver()->setTimeouts(['implicit' => 0]);
```
stof, line 38: Should we test the other kind of timeouts as well ?

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1345 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function isStarted(): bool
     {
-        return false;
+        return $this->started;
```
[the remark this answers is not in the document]
stof: I also suggest creating a private helper

```php
private function getWebDriver(): WebDriver
{
    if ($this->webDriver === null) {
        throw new DriverException('The driver is not started.');
    }

    return $this->webDriver;
}
```

Using this private helper in methods would allow to provide a better feedback than a fatal error when calling a method on a non-started driver, and it will help once we enable static analysis (for now, I added phpstan only on the Mink repo, not yet on drivers, but that's a next step).

stof (changes requested):

src/WebdriverClassicDriver.php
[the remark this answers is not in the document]
stof: We generally even keep the arguments on a single line.

resources/get-syn.php
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Simple script that updates syn.js by downloading it from NPM through Unpkg and attaches a file
+ * header comment. To switch to a different Syn version, just update $version variable appropriately.
+ *
+ * @codeCoverageIgnore
+ */
+
+$version = '0.15.0';
+
+if ('cli' !== PHP_SAPI && 'phpdbg' !== PHP_SAPI) {
```
oleg-andreyev, line 12: I'd rather remove syn.js, its not maintained

src/WebdriverClassicDriver.php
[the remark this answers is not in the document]
uuf6429: Ever since arguments (and especially, constructors' arguments) in PHP got fancier (typing, attributes and so on), I found it easier to have them on separate lines.
Additionally, it also improves readability of git 'blame' history.

.github/workflows/ci.yml
```diff
@@ -23,39 +23,61 @@ jobs:
...
+          DRIVER_URL: http://172.18.0.2:4444/wd/hub
+          WEB_FIXTURES_HOST: http://host.docker.internal:8002
+          DRIVER_MACHINE_BASE_PATH: /fixtures/
         run: |
-          vendor/bin/phpunit -v --coverage-clover=coverage.xml
+          vendor/bin/phpunit -v --coverage-clover=coverage.xml --colors=always --testdox
 
       - name: Upload coverage
         uses: codecov/codecov-action@v3
+        if: ${{ !env.ACT }}
```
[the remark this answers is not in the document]
uuf6429: https://github.com/nektos/act#skipping-steps

`act` is a tool to run github pipelines locally (very useful for testing the pipeline)

.github/workflows/ci.yml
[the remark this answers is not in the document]
uuf6429: Actually, I remember why I went this way - it's easier than trying to find the browser/selenium version when running tests.

I can go with the 2nd approach, and run some JS to find the browser type and version.

resources/get-syn.php
uuf6429: The selenium action api could definitely replace it. Of course we need to also consider older selenium as well.

At this point, if we decide to keep it, at least it's better to have this simple PHP script that explains/documents where syn came from.

I noticed you didn't take this approach, maybe you could try implementing your changes when this PR is merged (but before a first release)?

stof: Well, anything that can be done using the webdriver action API should be done with it. Places where Selenium2Driver is using syn.js are often places where it simulates half of the behavior but not all of it (for instance, the `focus` method simulates a focus event, but it might not actually focus the field).

.github/workflows/ci.yml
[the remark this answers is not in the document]
stof: or just pass en env variable in the CI job

.github/workflows/ci.yml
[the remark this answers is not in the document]
stof: I suggest adding a comment

uuf6429:
@stof I'm guessing this PR's jobs are failing due to this deprecation notice:
image
(there are other deprecation notices)

This is one of the things that confused me originally - why does the bridge fail for my branch, but passes for MinkSelenium2Driver? As far as I am aware, the relevant code is the same and both projects are not defining any bridge config.

Additionally, I cannot even fix that deprecation since `mixed` is available since PHP 8.

.github/workflows/ci.yml
```diff
@@ -26,36 +27,55 @@ jobs:
...
+          SELENIUM_IMAGE=selenium/standalone-${{ matrix.browser }}:${{ matrix.selenium }} docker compose up --no-color &> ./logs/selenium.log &
+
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
           coverage: "xdebug"
           php-version: "${{ matrix.php }}"
           ini-file: development
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
aik099, line 50: Why this is needed?

phpunit.xml.dist
```diff
@@ -18,15 +19,16 @@
...
-        <!--<server name="WEB_FIXTURES_BROWSER" value="firefox"/>-->
+
         <!-- where driver will connect to -->
         <server name="DRIVER_URL" value="http://localhost:4444/wd/hub"/>
+
         <!-- where DocumentRoot of 'Test Machine' is mounted to on 'Driver Machine' (only if these are 2 different machines) -->
-        <!--server name="DRIVER_MACHINE_BASE_PATH" value="" /-->
-        <!--server name="TEST_MACHINE_BASE_PATH" value="" /-->
+        <server name="DRIVER_MACHINE_BASE_PATH" value="/fixtures/"/>
     </php>
```
aik099, line 32: Defining ENV vars with Docker defaults prevents a developer from running a test suite with a locally started (not Docker-based) Selenium server.

I recommend replacing with code with the code of MinkSelenium2Driver.

.github/workflows/ci.yml
uuf6429: From what I understand, that disables github rate limiting anonymous Composer calls/downloads.

It's needed in our case due to the large number of parallel jobs. Later we could also enable composer caching though.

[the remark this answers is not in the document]
stof: and zip downloads should not be rate-limited on github (at least not with the low rate limit of API calls) so is this actually needed ?

tests/Custom/TimeoutTest.php
```diff
@@ -0,0 +1,59 @@
...
+
+        // Reset the array of timeouts to avoid impacting other tests
+        $session->getDriver()->setTimeouts([]);
+
+        parent::resetSessions();
+    }
+
+    public function testInvalidTimeoutSettingThrowsException(): void
+    {
+        $this->expectException(DriverException::class);
```
stof, line 30: this should be placed after the `start()` call, to ensure that the exception happens in `setTimouts`

tests/Custom/ChromeDesiredCapabilitiesTest.php
```diff
@@ -0,0 +1,71 @@
...
+        $driver = new WebdriverClassicDriver('chrome', $caps);
+        $this->assertNotEmpty($driver->getDesiredCapabilities(), 'desiredCapabilities empty');
+        $this->assertIsArray($driver->getDesiredCapabilities());
+        $this->assertEquals($caps, $driver->getDesiredCapabilities());
+    }
+
+    public function testSetDesiredCapabilities(): void
+    {
+        $this->expectException(DriverException::class);
+        $this->expectExceptionMessage('Unable to set desiredCapabilities, the session has already started');
```
stof, line 55: those should be moved just before the call in which we except the exception, so that the setup code before it does not expect exceptions (and so exceptions there mark the test as errored)

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1390 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     * @throws DriverException
+     */
+    public function switchToIFrame($name = null): void
+    {
+        $frameQuery = $name;
+        if ($name && is_string($name) && $this->getWebDriver()->isW3cCompliant()) {
```
stof, line 222: this `is_string($name)` is useless. As the argument is `string|null`, it will always be true here.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1390 @@
...
+     * @throws DriverException
+     */
+    public function forward(): void
+    {
+        $this->getWebDriver()->navigate()->forward();
+    }
+
+    /**
+     * {@inheritdoc}
+     * @throws DriverException
```
stof, line 191: please remove all those phpdoc containing only

```php
    /**
     * {@inheritdoc}
     * @throws DriverException
     */
```

or 

```php
    /**
     * {@inheritdoc}
     */
```

.github/workflows/ci.yml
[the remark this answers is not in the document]
uuf6429: Actually, I followed the instructions here: https://github.com/shivammathur/setup-php#:~:text=To%20avoid%20this%2C%20it%20is%20recommended%20to%20provide%20a%20GitHub%20OAuth%20token. That secret is also [the same token generated by github](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow).

I'm not sure it is needed anymore though, I'll remove it for now.

tests/Custom/TimeoutTest.php
uuf6429: Agreed... these tests were copied from MinkSelenium2Driver; they could probably be improved.

.github/workflows/ci.yml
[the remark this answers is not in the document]
stof: ah, I forgot that GHA defines a default secret for the token as well.

stof:
@uuf6429 the DebugClassLoader triggers a deprecation only for classes that are not in a common namespace. Due to the way namespaces were defined for the older drivers (which is a legacy from the fact that they were initially in the main Mink repo and they kept their FQCN when splitting them), DebugClassLoader skips those deprecations for Selenium2Driver.
The reasoning for that is that for classes that have the same namespace, you will add the typehint in both at the same time (reusing the same namespace in separate composer packages is against the best practices).

Regarding not being able to fix it due to the min PHP version, if you read the deprecation message again, you will see that DebugClassLoader will also stop reporting the deprecation if you have an explicit `@return mixed` in your phpdoc (not relying on the parent phpdoc) as it will expect you to add that as a native type in a later major version.
an alternative would be to make this new driver require PHP 8+ instead of 7.4+. When at look at the installation stats of BrowserKitDriver (the only driver for which we made a new major version for which projects had to update their requirements instead of just benefiting from semver ranges), the v2 has 90.5% of PHP 8+ for the versions used. To me, it seems likely that projects that would do the effort to migrate to a new driver will be on a maintained PHP version (which is not the case of PHP 7.4)

uuf6429:
After the latest iteration, most of the tests are passing 💪

One particular failure is from chrome-selenium-2 `\Behat\Mink\Tests\Driver\Js\WindowTest::testResizeWindow` test.
When the test is run in isolation it always works, but when the whole class is run, it's always failing. 

Adding `$this->getSession()->restart()` to the beginning of the test seems to fix it.

If I understand the situation correctly, the previous tests are opening (but not closing) some popups, and in some case, the popup is left as the main window. When this test runs and tries to resize the popup, it's not able to do that and therefore it fails the size check. (at least this is my guess)

1. is it ok to just restart the session for this test? it feels like it's only hiding the main problem
2. is it normal that there leftover popups from previous tests?
3. session/driver `reset()` does not solve the problem, but I also noticed that the DriverInterface doesn't say that existing windows should be closed

src/WebdriverClassicDriver.php
uuf6429: I removed `@throws DriverException` in all such cases except for reset(), where throwing is new due to documented undefined behaviour.

Keeping the inheritdoc ones is useful since the IDE renders the resolved documentation nicely:
image
what do you think?

tests/Custom/TimeoutTest.php
```diff
@@ -48,12 +51,18 @@ public function testShortTimeoutDoesNotWaitForElementToAppear(): void
...
         $this->getSession()->visit($this->pathTo('/js_test.html'));
         $this->findById('waitable')->click();
         $element = $this->getSession()->getPage()->find('css', '#waitable > div');
 
         $this->assertNotNull($element);
     }
+
+    private function getDriver(): WebdriverClassicDriver
+    {
+        /** @phpstan-ignore-next-line */
```
stof, line 65: you should instead add an assert, as I did in Selenium2Driver. This is better than ignoring errors (in development mode, it will actually fail if the invariant is wrong)

src/WebdriverClassicDriver.php
```diff
@@ -1386,5 +1389,10 @@ private function jsonEncode($value, string $action, string $field): string
         }
     }
 
+    private function wrapError(Throwable $e): Exception
+    {
+        return $e instanceof Exception ? $e : new ErrorException($e);
```
stof, line 1357: I think we should rather add support for any Throwable as the previous exception in Mink's DriverException instead of having this wrapping (after all, the only thing it does with it is passing it to the parent constructor).

stof:
Closing the main window should not be needed during reset (after all, the whole point of reset is to be faster than a restart by avoiding to close and reopen the browser).
but looking at the way the initial window is handled in the driver implementation, `reset` should probably switch back to it if the current window is not that one.

src/WebdriverClassicDriver.php
stof: Inheriting the description should already be the behavior of phpdoc when there is no phpdoc on the element.
When *using* the class, PHPStorm might even maybe do it properly (when opening a file, it shows only what this file contains apparently).
The `{@inheritdoc}` tag is only needed when you want to include the inherited description as part of a longer one in the child class (one that adds additional info)

src/WebdriverClassicDriver.php
uuf6429: I think that should be doable without requiring a version since it concerns constructors, right?

If so, could you please take care of it?

tests/Custom/TimeoutTest.php
stof: I mean as I did in https://github.com/minkphp/MinkSelenium2Driver/pull/369 which is not merged yet.

src/WebdriverClassicDriver.php
stof: See https://github.com/minkphp/Mink/pull/855

composer.json
```diff
@@ -14,14 +14,17 @@
     ],
     "require": {
         "php": ">=7.4",
-        "behat/mink": "^1.9@dev"
+        "ext-json": "*",
+        "behat/mink": "^1.9@dev",
```
stof, line 18: this should be changed to `^1.11@dev` to support passing Throwable in the previous exception.

tests/Custom/TimeoutTest.php
uuf6429: I'm not sure I see the difference - in both cases an error is triggered (unless assertions have been disabled), and afaik for assertions the failure message won't be very descriptive, unless I'm missing something?

stof: the difference is that `@phpstan-ignore-next-line` ignores *any* error happening on a next line.
An assertion will then make phpstan aware that we deal with a specific kind of value and it will continue its analysis based on that, and will still report other errors if we do them.
and this is not really a case where we expect the assertion to fail (due to the way we setup the testsuite, we know which driver we have)

src/WebdriverClassicDriver.php
uuf6429: PHPDoc inheritance in PHPStorm works well (for users of the class or interface), however when you open this class directly, it won't show documentation of those methods in the same file.

In other words, without inheritdoc we would miss things like exceptions thrown or arguments typehinting from the interface (but this won't impact the end user).

If you still have a strong opinion on this, I can remove it - presumably not a lot of people would be opening this file directly.

.github/workflows/ci.yml
uuf6429: As mentioned below, I removed it for now.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1361 @@
...
+            $name = '';
+        }
+
+        $formId = $element->getAttribute('form');
+        if ($formId === true) {
+            $formId = '';
+        }
+
+        try {
+            $escapedName = $this->xpathEscaper->escapeLiteral($name);
```
stof, line 1240: Can't you use the `WebDriverRadios` helper class instead ?

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1361 @@
...
+    }
+
+    /**
+     * @throws DriverException
+     */
+    private function selectOptionOnElement(RemoteWebElement $element, string $value, bool $multiple = false): void
+    {
+        $escapedValue = $this->xpathEscaper->escapeLiteral($value);
+        // The value of an option is the normalized version of its text when it has no value attribute
+        $optionQuery = sprintf(
```
stof, line 1272: I suggest using `WebDriverSelect` instead of redoing a similar logic, if possible.

src/WebdriverClassicDriver.php
stof: I suggest avoiding to duplicate the parent phpdoc. It avoids having to update it when we update it in the parent class.

Contributors working on the class (which will be at most 3 devs probably, including you and me) can navigate to the interface in phpstorm for the case we need to see it (and maybe in the future PHPStorm will add this).

stof:
@uuf6429 I just merged a few new tests in the driver-testsuite about error handling for invalid usages of `setValue` (passing the wrong type of value depending on the kind of fields). Those are the ones that failed in the latest run of the CI.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+        } catch (NoSuchCookieException $e) {
+            $result = null;
+        }
+        if ($result === null) {
+            return null;
+        }
+
+        $result = $result->getValue();
+        if ($result === null) {
+            return null;
```
stof, line 250: this condition looks like dead code. The php-webdriver defines the return type of `getValue` as being `string`

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+            $name = self::W3C_WINDOW_HANDLE_PREFIX . $this->getWebDriver()->getWindowHandle();
+        }
+
+        return $name;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function findElementXpaths(
```
stof, line 308: should be protected to match the parent visibility

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+        $elementName = strtolower($element->getTagName() ?? '');
+        $elementType = strtolower((string)$element->getAttribute('type'));
+
+        // Getting the value of a checkbox returns its value if selected.
+        if ($elementName === 'input' && $elementType === 'checkbox') {
+            return $element->isSelected() ? $element->getAttribute('value') : null;
+        }
+
+        if ($elementName === 'input' && $elementType === 'radio') {
+            $script = <<<JS
```
stof, line 393: I suggest implementing it using `WebDriverRadios` instead

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+                return value;
+                JS;
+
+            return $this->executeJsOnElement($element, $script);
+        }
+
+        // Using $element->attribute('value') on a select only returns the first selected option
+        // even when it is a multiple select, so a custom retrieval is needed.
+        if ($elementName === 'select' && $element->getAttribute('multiple')) {
+            $script = <<<JS
```
stof, line 419: I suggest using `WebDriverSelect`

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+
+                return value;
+                JS;
+
+            return $this->executeJsOnElement($element, $script);
+        }
+
+        // use textarea.value rather than textarea.getAttribute(value) for chrome 91+ support
+        if ($elementName === 'textarea') {
+            $script = <<<JS
```
stof, line 437: this should actually use `$element->getDomProperty('value')` on W3C-compliant Webdriver (probably also for inputs btw)

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getAttribute(
+        #[Language('XPath')]
+        $xpath,
+        $name
+    ): ?string {
+        $escapedName = $this->jsonEncode($name, 'get attribute', 'attribute name');
+        $script = "return arguments[0].getAttribute($escapedName)";
```
stof, line 371: can't we use `$element->getAttribute($name)`, at least for W3C-compliant sessions instead of this need for escaping ? JSONWire had a weird behavior where `getAttribute` wads not really getting the attribute, but this does not seem to happen in W3C webdriver.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function executeScript(
+        #[Language('JavaScript')]
+        $script
+    ): void {
+        if (preg_match('/^function[\s(]/', $script ?? '')) {
```
stof, line 761: `?? ''` is useless. Null values are not supported in this method for the argument

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+            $script = preg_replace('/;$/', '', $script ?? '');
```
stof, line 762: I suggest doing `$script = rtrim($script, ';')` as a simpler implementation.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+
+    /**
+     * {@inheritdoc}
+     * @return mixed
+     */
+    public function evaluateScript(
+        #[Language('JavaScript')]
+        $script
+    ) {
+        if (strncmp(ltrim((string)$script), 'return ', 7) !== 0) {
```
stof, line 777: the string cast is useless

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+
+        throw new DriverException('Driver has not been started');
+    }
+
+    /**
+     * Detect and assign appropriate browser capabilities
+     *
+     * @see https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities
+     */
+    private function initCapabilities(array $desiredCapabilities = []): DesiredCapabilities
```
stof, line 978: to me, there is no reason to make this argument optional in the private method

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1328 @@
...
+    {
+        // Build base capabilities
+        $browserName = $this->browserName;
+        if ($browserName && method_exists(DesiredCapabilities::class, $browserName)) {
```
stof, line 982: the fact that the method exists does not mean that it is a static factory method

stof:
As https://github.com/minkphp/Mink/pull/856 has been merged, you can now use native parameter types in the driver public methods (except for `$value` in `setValue`, `$user` in `setBasicAuth` and `$char` in `key*` methods as those arguments would require using a union type and so they don't have a type yet in Mink). This requires making Mink 1.11 the minimum requirement, but this is already the case in this PR.

src/WebdriverClassicDriver.php
uuf6429: Seems that the only way is to keep a list of such methods. I realised the current code is wrong anyway, since it will look for a method called 'internet explorer' for example.

stof:
Selenium2Driver does not support setting basic auth. I think this new driver cannot support it either (unless Webdriver added a feature to allow that since that time, but I don't think so).

tests/Custom/ChromeDesiredCapabilitiesTest.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+namespace Mink\WebdriverClassDriver\Tests\Custom;
+
+use Behat\Mink\Exception\DriverException;
+use Behat\Mink\Tests\Driver\TestCase;
+use Mink\WebdriverClassDriver\WebdriverClassicDriver;
+
+class ChromeDesiredCapabilitiesTest extends TestCase
```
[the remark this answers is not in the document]
uuf6429: When I think about it, this capability stuff is still quite overcomplicated and a bit stupid:

- keep in mind setting/getting capabilities is not part of the driver interface
- restarting a driver to use new capabilities is as good as recreating the driver object with such capabilities, so might as well pass capabilities a a constructor dependency
- getting back the capabilities at the moment gives the wrong impression; some capabilities can be filtered out at the php-webdriver or even selenium level, and we won't know this from `getDesiredCapabilities()`
- as far as I am aware, there's no further use for getting the capabilities, so might as well leave that functionality out

So in my opinion, we could keep initCapabilities() and remove set/get one.

With that change, the aforementioned tests become completely redundant.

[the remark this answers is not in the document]
uuf6429: I've applied that change now, let me know if you have any objections.

.github/workflows/ci.yml
```diff
@@ -23,39 +23,61 @@ jobs:
...
       - name: Upload coverage
         uses: codecov/codecov-action@v3
+        if: ${{ !env.ACT }}
         with:
           files: coverage.xml
 
       - name: Archive logs artifacts
-        if: ${{ failure() }}
         uses: actions/upload-artifact@v3
+        if: always()
```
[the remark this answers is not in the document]
uuf6429: right...I'm not sure anymore why, guess it's some copy&paste mistake 😅

tests/WebdriverClassicConfig.php
```diff
@@ -2,31 +2,56 @@
...
 
 class WebdriverClassicConfig extends AbstractConfig
 {
-    public static function getInstance()
+    public static function getInstance(): self
     {
         return new self();
     }
 
-    public function createDriver(): DriverInterface
```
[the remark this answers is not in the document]
stof: Well, a previous version of the PR was remove the `DriverInterface` return type. But this has already been fixed

tests/Custom/ChromeDesiredCapabilitiesTest.php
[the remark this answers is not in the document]
stof: We should still have tests covering the usage of custom capabilities though (not sure how this can be tested)

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1187 @@
...
+    private string $webDriverHost;
+
+    private ?string $initialWindowName = null;
+
+    /**
+     * @throws DriverException
+     */
+    public function __construct(
+        string $browserName = self::DEFAULT_BROWSER,
+        array $desiredCapabilities = [],
```
stof, line 86: should we support passing a DesiredCapabilities object directly ?

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1187 @@
...
+
+            case 'input':
+                $elementType = strtolower((string)$element->getAttribute('type'));
+                switch ($elementType) {
+                    case 'submit':
+                    case 'image':
+                    case 'button':
+                    case 'reset':
+                        $message = 'Cannot set value on element with XPath "%s" as it is not a select, textarea or textbox';
+                        throw new DriverException(sprintf($message, $xpath));
```
stof, line 398: The fact that this line is not covered by tests tells me that we are missing some tests in the driver-testsuite ensuring that those input types cannot be used with `setValue`

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1187 @@
...
+
+                    case 'color':
+                        if (!is_string($value)) {
+                            throw new DriverException('Color value must be a string');
```
stof, line 402: This case should be covered by tests in the HTML5Test of the driver testsuite (similar to the tests I added recently in SelectTest and RadioTest for similar bad usages).

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1187 @@
...
+
+            case 'select':
+                if (is_array($value)) {
+                    $this->deselectAllOptions($element);
+                    foreach ($value as $option) {
+                        $this->selectOptionOnElement($element, $option, true);
+                    }
+                    return;
+                }
+                $this->selectOptionOnElement($element, (string)$value);
```
stof, line 387: This should throw an exception for boolean values instead of casting them to string.

src/WebdriverClassicDriver.php
uuf6429: Probably not:
- that class is not part of the public interface of this library
  - the right way would be to provide our own (perhaps a subclass of the original)
- we wouldn't be able to add in the default browser-specific capabilities

tests/Custom/ChromeDesiredCapabilitiesTest.php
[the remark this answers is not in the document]
uuf6429: Makes sense.

One possibility is to use some well-known capability, such as [setWindowRect](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Commands/SetWindowRect) and then checking the window size, maybe?

tests/Custom/WindowNameTest.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    protected function tearDown(): void
+    {
+        $this->getSession()->stop();
+
+        parent::tearDown();
+    }
+
+    public function testWindowNames(): void
```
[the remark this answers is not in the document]
uuf6429: Done: https://github.com/minkphp/driver-testsuite/pull/76

tests/Custom/ChromeDesiredCapabilitiesTest.php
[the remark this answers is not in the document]
uuf6429: I couldn't find a sane way either - the only options I see are:
- testing with `acceptInsecureCerts` and some intentionally insecure url (ofc one needs to test the opposite as well)
- some fancy browser-specific capability

In both cases, the test's scope would be limited (in the first case, that capablity might not be available in old browsers)

src/WebdriverClassicDriver.php
uuf6429: Also, partly related: apparently the concept of "desired" and "required" capabilities is now deprecated, so we should also avoid that naming within the driver and especially the public interface (I changed this already).

src/WebdriverClassicDriver.php
uuf6429: Done: https://github.com/minkphp/driver-testsuite/pull/77

src/WebdriverClassicDriver.php
uuf6429: Done: https://github.com/minkphp/driver-testsuite/pull/78

uuf6429:
@minkphp/webdriver-classic hey, I created separate tasks/issues for the rest of the suggestions.

What's next? Is there anything else missing? I think the main point is to get the rest of the new tests merged.

phpunit.xml.dist
```diff
@@ -18,15 +19,18 @@
...
+
         <!-- where driver will connect to -->
         <server name="DRIVER_URL" value="http://localhost:4444/wd/hub"/>
+
         <!-- where DocumentRoot of 'Test Machine' is mounted to on 'Driver Machine' (only if these are 2 different machines) -->
-        <!--server name="DRIVER_MACHINE_BASE_PATH" value="" /-->
-        <!--server name="TEST_MACHINE_BASE_PATH" value="" /-->
+        <server name="DRIVER_MACHINE_BASE_PATH" value="/fixtures/"/>
+
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak"/>
```
aik099, line 33: Changing the defaults of this section would result in the developer's inability to run the test suite with a locally started Selenium server.

aik099:
> What's next? Is there anything else missing? I think the main point is to get the rest of the new tests merged.

@uuf6429 , I see, that some discussions on PR code are still unresolved. Please address them.

phpunit.xml.dist
uuf6429: I replied to this point already, but I guess it got somehow lost:
https://github.com/minkphp/webdriver-classic-driver/pull/1#discussion_r1229075107

That still stands, the recommended test setup is with docker, therefore the defaults are tailored for docker. In case developers want to customize this, they can easily create their own `phpunit.xml` (that won't be committed with the project).

uuf6429:
> > What's next? Is there anything else missing? I think the main point is to get the rest of the new tests merged.
> 
> @uuf6429 , I see, that some discussions on PR code are still unresolved. Please address them.

@aik099 I replied to all of them. From my perspective the main remaining point is to have all the [relevant PRs in WebDriverTests merged](https://github.com/minkphp/driver-testsuite/pulls) to be sure the driver is up to spec.
There are some additional tasks, which I opened new GH issues for and which won't make it into the first alpha version.

If all your concerns have been addressed, please approve the PR instead of waiting for someone else to make their first move.

tests/WebdriverClassicConfig.php
```diff
@@ -15,22 +21,62 @@ public static function getInstance(): self
...
 
+    /**
+     * {@inheritdoc}
+     */
     protected function supportsCss(): bool
     {
         return true;
     }
+
+    private function isXvfb(): bool
```
oleg-andreyev, line 72: misleading method name. why you need it?

src/WebdriverClassicDriver.php
oleg-andreyev: take a look at https://github.com/oleg-andreyev/MinkPhpWebDriver/blob/058c254b04f9586d16fccc82ae2bc0c5e88c402f/src/WebDriver.php#L540-L558

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1195 @@
...
+            'platform' => 'ANY',
+            'name' => 'Behat Test',
+            'deviceOrientation' => 'landscape',
+            'deviceType' => 'desktop',
+        ],
+
+        'chrome' => [
+            'goog:chromeOptions' => [
+                // disable "Chrome is being controlled.." notification bar
+                'excludeSwitches' => ['enable-automation'],
```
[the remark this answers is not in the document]
uuf6429: Not sure what you mean...the comment describes exactly what it does...removing the notification bar means there's more rendering space and less unexpected UI potentially interfering with tests.

src/WebdriverClassicDriver.php
```diff
@@ -1,21 +1,1195 @@
...
+
+    public function stop(): void
+    {
+        if (!$this->webDriver) {
+            throw new DriverException('Driver has not been started');
+        }
+
+        try {
+            $this->getWebDriver()->quit();
+            $this->webDriver = null;
```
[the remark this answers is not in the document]
uuf6429: Makes sense!

tests/WebdriverClassicConfig.php
uuf6429: > misleading method name
It's not exactly misleading - it's what we are actually intending there by checking the github action env var.

That check was originally implemented in MinkSelenium2Driver:
https://github.com/minkphp/MinkSelenium2Driver/blob/master/tests/Selenium2Config.php#L38

After trial and error, I found what it was trying to do, hence the name I gave it. Maybe we can find a better name though.

tests/WebdriverClassicConfig.php
```diff
@@ -15,22 +21,62 @@ public static function getInstance(): self
...
+    {
+        return getenv('GITHUB_ACTIONS') === 'true';
+    }
+
+    private function isOldChrome(): bool
```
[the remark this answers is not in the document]
uuf6429: Selenium 3 images are bundling an old version of chrome (which is causing the root problem).

Unfortunately, from this class's perspective, there's no way of finding the real chrome version.

tests/WebdriverClassicConfig.php
uuf6429: Also, it gets its name from here: https://github.com/minkphp/webdriver-classic-driver/pull/1/files/d385eed7f7ec955f3706d765003a5907beb85509#diff-db0e666a5299190f519b12062eb766dccd38dc00a2c6813ec997ed4e6e29b186R45

src/WebdriverClassicDriver.php
uuf6429: @oleg-andreyev I thought about that as well and decided against it because:
1. that takes at least 3 requests, mine only takes one
2. it requires explaining why the extra logic is needed
3. this approach is backward compatible (in older selenium)

src/WebdriverClassicDriver.php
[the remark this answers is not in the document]
uuf6429: The situation for Edge is similar, except that for some weird reason, it refuses to even start when with any settings (but that seems to be something else).

src/WebdriverClassicDriver.php
```diff
@@ -11,11 +11,1194 @@
...
+
+    private string $webDriverHost;
+
+    private ?string $initialWindowName = null;
+
+    /**
+     * @param string $browserName One of 'edge', 'firefox', 'chrome' or any one of {@see WebDriverBrowserType} constants.
+     */
+    public function __construct(
+        string $browserName = self::DEFAULT_BROWSER,
```
uuf6429, line 89: @stof recently I found that the proper browser name for MSEdge is "MicrosoftEdge".
For some reason, "edge" used to work before, but not anymore.
At the moment, we use the browser name from the docker image suffix, which eventually triggered the problem I just mentioned.

However, this got me thinking, at the moment there aren't any restrictions or suggestions for the browser name from the mink side, which is far from ideal especially since we have extra (default) initialization logic based on the browser name.

On the other hand, I'm not sure it's a good idea to tell the dev to use constants from a 3rd party library (what I sort of the in the phpdoc). I took the liberty to say exactly which browsers are definitely supported in the PHPDoc, which is effectively the keys from `BROWSER_NAME_ALIAS_MAP`. Another approach could be to define our own constants, or provide them in a new class that extends `WebDriverBrowserType`.

What do you think?

src/WebdriverClassicDriver.php
```diff
@@ -11,11 +11,1194 @@
...
+
+            case WebDriverBrowserType::SAFARI:
+                return DesiredCapabilities::safari();
+
+            case WebDriverBrowserType::OPERA:
+                return DesiredCapabilities::opera();
+
+            case WebDriverBrowserType::MICROSOFT_EDGE:
+                return DesiredCapabilities::microsoftEdge()
+                    ->setCapability(WebDriverCapabilityType::PLATFORM, WebDriverPlatform::ANY);
```
uuf6429, line 796: See: https://github.com/php-webdriver/php-webdriver/issues/1076

src/WebdriverClassicDriver.php
```diff
@@ -11,11 +11,1194 @@
...
+        }
+    }
+
+    /**
+     * @param mixed $value
+     * @throws DriverException
+     * @phpstan-assert string $value
+     * @todo When switching to PHP 8, this can be replaced with `is_string(..) or throw new DriverException(..);`
+     */
+    private function assertString($value, string $message): void
```
stof, line 1151: As this method is used only in contexts that do not require an expression (which would indeed not be possible in PHP 7.4), I suggest inlining it.
There is no reason to add this method with a TODO asking to replace it with an expression when we can already replace it.

src/WebdriverClassicDriver.php
```diff
@@ -11,11 +11,1194 @@
...
+        }
+    }
+
+    /**
+     * @param mixed $value
+     * @throws DriverException
+     * @phpstan-assert bool $value
+     * @todo When switching to PHP 8, this can be replaced with `is_bool(..) or throw new DriverException(..);`
+     */
+    private function assertBool($value, string $message): void
```
stof, line 1164: this one is used only once in a context allowing statements, so I would inline it.

.github/workflows/ci.yml
```diff
@@ -57,19 +68,30 @@ jobs:
         run: |
           composer update --no-interaction --prefer-dist --ansi --no-progress
 
+      - name: Wait for selenium to start
+        run: |
+          curl --head -X GET --silent --show-error --retry 60 --retry-connrefused --retry-delay 1 --max-time 10 http://172.18.0.2:4444
```
stof, line 73: what is this hardcoded IP here ?

stof (commented):
Regarding the usage of syn.js, I'd rather keep the initial implementation free of it, marking methods as unsupported and skipping tests if needed. 

syn.js is quite a bad fit as it does not actually simulate the actual browser events. See https://github.com/minkphp/driver-testsuite/issues/36#issuecomment-559391149

Note that we might have to deprecate the existing keypress, keyup and keydown methods as they don't reflect reality: an actual browser has no way to interact in a way that triggers only one of those events.

.github/workflows/ci.yml
uuf6429: It's the same one as `DRIVER_URL` a few lines below.

It references docker on the host machine (we are running selenium in docker while running tests on the host machine).
Normally, one would use 127.0.0.1/localhost or even 0.0.0.0, but it doesn't seem to work in github.
When I have some time, I'll see if I can come up with an alternative.

stof: is this IP guaranteed to stay the same ?

Otherwise, it might be better to make docker expose the port on localhost

stof: btw, looking at the docker-compose file, it looks like you _already_ expose the port.

uuf6429: > btw, looking at the docker-compose file, it looks like you already expose the port.

Yes, that's what I meant by

> Normally, one would use 127.0.0.1/localhost or even 0.0.0.0, but it doesn't seem to work in github.

Somehow github doesn't bind the port to localhost, but it's accessible from that ip I used.

Anyway, I'll try a few approaches here: https://github.com/uuf6429/webdriver-classic-driver/pull/1
If I can't manage, I'll at least try to find some reference on how I got that ip and add it as a comment.

uuf6429: Fixed!

stof:
@uuf6429 any chance to fix the CI jobs that are failing ?

uuf6429:
> @uuf6429 any chance to fix the CI jobs that are failing ?

@stof if I remember correctly, they're blocked by https://github.com/minkphp/driver-testsuite/pull/84, https://github.com/minkphp/driver-testsuite/pull/75 (maybe?) and https://github.com/uuf6429/webdriver-classic-driver/pull/2 (eventually, I guess).

PR #6, opened 2023-12-09
Remove syn (closes #5)

### LINKED ISSUE #5: Replace syn.js
If possible, use selenium action api for all cases.

If not, consider using a syn alternative (or if to complex, depend on syn) for old selenium.

uuf6429:


tests/WebdriverClassicConfig.php
```diff
@@ -53,9 +52,16 @@ public function skipMessage($testCase, $test): ?string
             case $testCase === StatusCodeTest::class:
                 return 'Checking status code is not supported.';
 
-            case $testCase === EventsTest::class && $test === 'testKeyboardEvents' && $this->isOldChrome():
+            case [$testCase, $test] === [EventsTest::class, 'testKeyboardEvents'] && $this->isOldChrome():
                 return 'Old Chrome does not allow triggering events.';
 
+            case [$testCase, $test] === [EventsTest::class, 'testBlur']:
+            case [$testCase, $test] === [EventsTest::class, 'testFocus']:
+                return 'Focus/blur are not supported anymore.';
```
stof, line 60: I'm wondering whether this could be supported by executing `element.focus()` (or `element.blur()`) in JS.

src/WebdriverClassicDriver.php
```diff
@@ -1128,19 +1034,6 @@ private function ensureInputType(
...
-     * @throws DriverException
-     */
-    private function trigger(
-        #[Language('XPath')]
-        string $xpath,
-        string $event,
-        #[Language('JSON')]
-        string $options = '{}'
-    ): void {
-        $this->withSyn()->executeJsOnXpath($xpath, "window.syn.trigger(arguments[0], '$event', $options)");
```
aik099, line 1141: Why removing it completely. Can't fire events using native JavaScript like so:

```js
$0.dispatchEvent(new Event('blur'));
```

stof: That's the whole point: we should not only fire a fake event. That's not the intended behavior of this method. `blur` is expected to actually remove the focus from the element.

But this case should probably be doable (see my previous review)

tests/WebdriverClassicConfig.php
uuf6429: I tried it locally and they seem to work, however the [testBlur](https://github.com/minkphp/driver-testsuite/blob/f9d36b63480ce013d0eecc2fc0ce962ca956dff2/tests/Js/EventsTest.php#L66) test fails.

It kinda makes sense too, because you cannot trigger blur on a non-focused element.
If I add `$focusBlurDetector->focus();` before `$focusBlurDetector->blur();` then the test passes.

src/WebdriverClassicDriver.php
uuf6429: > Why removing it completely? Can't fire events using native JavaScript like so:
> 
> ```js
> $0.dispatchEvent(new Event('blur'));
> ```

@aik099 

I've checked the [original PR](https://github.com/uuf6429/webdriver-classic-driver/pull/2#issuecomment-1742088874) and the reason (from my side) was that end users cannot directly focus or blur elements. They perform actions that focus/blur as a side-effect. For example, when you "tab" between fields or "click" fields, you are firstly interacting with the keyboard or mouse. There isn't a way for a user to "just" focus or blur something.

And since we were removing syn and some related events, it made sense to me to also disable this functionality (if mink also eventually goes in that direction).

But I don't have a strong opinion on this, I'm also fine adding it back (as I just did).

src/WebdriverClassicDriver.php
```diff
@@ -455,8 +455,6 @@ public function setValue(
         } catch (\Throwable $e) {
             throw new DriverException("Cannot set $widgetType value: {$e->getMessage()}", 0, $e);
         }
-
-        $this->trigger($xpath, 'blur');
```
uuf6429, line 459: Other than the `testBlur` failure, there are 4 other failures caused by removing this line.

I already described the problem [here](https://github.com/uuf6429/webdriver-classic-driver/pull/2#issuecomment-1742057992). I could bring back `$this->blur($xpath)` here, but I believe this behaviour (and the failing tests) are wrong. Please let's decide on this since it's unclear for me how Mink should behave in this case. @stof

aik099: ## Here is what I remember (it might not be too detailed, but you can look it up via Git Blame in Selenium2Driver) behind the `blur` event firing from the `setValue` call:

1. we're sending the TAB key after a typed text to cause the native `change` event to be fired (if the value was actually changed)
2. then there was a bug in Chrome 53 that caused TAB to be displayed in the input
3. the above problem was fixed by firing the `change` event manually
4. unconditional `change` event firing caused another problem, that in some cases the `change` event was fired twice
5. to fix that we've replaced the `change` event firing with the `blur` event firing that in turn, if the element is still in focus, would cause the `change` event to be fired
6. this made testing of the auto-complete controls impossible because the `blur` event caused auto-completed suggestions to disappear
7. the above-mentioned people found some ways of auto-complete control testing and changes in the Mink Driver weren't necessary

## Current Selenium2 Driver code for the `blur` firing

```php
// Remove the focus from the element if the field still has focus in
// order to trigger the change event. By doing this instead of simply
// triggering the change event for the given xpath we ensure that the
// change event will not be triggered twice for the same element if it
// has lost focus in the meanwhile. If the element has lost focus
// already then there is nothing to do as this will already have caused
// the triggering of the change event for that element.
$script = <<<JS
var node = {{ELEMENT}};
if (document.activeElement === node) {
  document.activeElement.blur();
}
JS;

// Cover case, when an element was removed from DOM after its value was
// changed (e.g. by a JavaScript of a SPA) and therefore can't be focused.
try {
    $this->executeJsOnElement($element, $script);
} catch (StaleElementReference $e) {
    // Do nothing because an element was already removed and therefore
    // blurring is not needed.
}
```

I'm not seeing that code in this driver anymore? Was it removed on purpose?

uuf6429: I've read your points several times and somehow it feels more and important not to have the old/selenium2driver behaviour.

1. Sending tab sounds like an unreasonable and unexpected behaviour, which is bound to cause problems like point 2
2. Points 3 to 5 sounds like a hack and further fighting with unexpected and non-standard behaviour
3. Point 6 ("triggering blur causes problems"), seems to contradict the comment referring to it, where you seem to say that point 6 "is a good thing". Otherwise I don't get what you meant there.
4. Regarding point 7, it is in my opinion the responsibility of the user or library vendor. Not autocomplete components behave in the same way, and it would be way too much work for us to maintain that for them. 

> I'm not seeing that code in this driver anymore? Was it removed on purpose?

We kept the old logic of triggering blur (before removing syn), but not checking if the element has been removed - so maybe we might be missing a test for that.
But anyway, I don't know why this logic is significant/required. I don't expect "change" or "blur" events after changing some types of fields - I don't see why the driver/mink needs to do that.

If the driver user specifically needs that, they can just call the blur() after setValue() themselves. Deciding for them means that we will unconditionally affect everyone else, including those people that don't want this non-standard behaviour.

src/WebdriverClassicDriver.php
uuf6429: This has been reimplemented back. Let's continue the other discussion in the other comment thread.

src/WebdriverClassicDriver.php
aik099: > I've read your points several times and somehow it feels more convincing not to have the old/selenium2driver behaviour.

All these 7 points above are just a history to bring you up to speed with Selenium2 driver development.

As a Mink maintainer, my main concern is to make library users happier and ensure, that the library is evolving in the right direction. Ensuring that all drivers behave the same way, even if the test suite doesn't report any inconsistencies.

> If the driver user specifically needs that, they can just call the blur() after setValue() themselves.

If we use raw Mink, then we probably can. I know, that the change/blur event firing thing isn't an expected behavior of the driver, but it's a good thing to have. 

Maybe tab/blur/change code integrated into the `setValue` method is a leftover from times, when Mink was used only inside a Behat, where scenarios were saying to change field value and not to change field value and fire a change event.

stof: Setting the value without a change event being fired also breaks all cases where you have JS code running on the change event. That's why our testsuite ensures that setting the value actually reports it to JS that the value has changed.

There was a discussion a few years ago (that has never been implemented) about introducing a new Mink API to write in a field (which would not remove the existing text first and would keep the focus in the field) to allow testing such autocomplete-based UI.

tests/WebdriverClassicConfig.php
stof: I merged the PR in the driver testsuite

src/WebdriverClassicDriver.php
aik099: > Setting the value without a change event being fired also breaks all cases where you have JS code running on the change event. That's why our testsuite ensures that setting the value actually reports it to JS that the value has changed.

Wow. I totally forgot that. What was the name of that test method? I wanted to make sure it's also checking, that only a single `change` event is being fired.

uuf6429: Let's continue the discussion here: https://github.com/minkphp/webdriver-classic-driver/issues/21#issuecomment-1992659987

PR #13, opened 2024-02-25
The README.md improvements

aik099:
Addressed issues:

* the driver usage example PHP code is missing -> added an example;
* the Selenium 2 version is used -> using Selenium 3 version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,12 +44,12 @@ composer require behat/mink mink/webdriver-classic-driver
 1. Start WebDriver
     1. If you have Docker installed, run
     ```bash
-    docker run -p 4444:4444 selenium/standalone-firefox:2.53.1
+    docker run -p 4444:4444 selenium/standalone-firefox:3.141.59
     ```
     2. If you do not have Docker, but you have Java
     ```bash
-    curl -L https://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.1.jar > selenium-server-standalone-2.53.1.jar
-    java -jar selenium-server-standalone-2.53.1.jar
+    curl -L https://selenium-release.storage.googleapis.com/3.141/selenium-server-standalone-3.141.59.jar > selenium-server-standalone-3.141.59.jar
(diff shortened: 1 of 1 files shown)


uuf6429 (approved):
Some side notes:
1. We could also update the example selenium version to 4, since this driver actually supports it.
2. I'm a fan of running README code in tests, to ensure that that code remains executable and up-to-date with our changes. If we can agree on this, I can create an issue to set that up later.

   In theory it would look like:

   ~~~markdown
   ```php
   echo 'Hello world';

   // Outputs:
   // Hello world
   ```
   ~~~

   ```php
   /**
    * @dataProvider readmeExamplesDataProvider
    */
   public function testThatExampleWorks(string $lang, string $input, string $output): void
   {
       switch ($lang) {
           case 'php':
               $this->expectOutputString($output);
               eval($input);
               return;

           // ...
       }
   }
   ```
   _(`readmeExamplesDataProvider` parses the readme and would return `['php', "echo 'Hello world';", 'Hello world']`)_


After you decide on those two points, feel free to merge this PR.

AlexSkrypnyk:
Could you please also add a section that would explain the difference between this project and `minkphp/MinkSelenium2Driver` because it is quite hard to understand what "classic" means exactly and what the roadmap looks like.

I'm coming from Drupal and Behat, and Mink driver is not directly used, so I'm just trying to wrap my head around what this classic driver will replace and whether that replacement be a drop-in or if we need to re-write some other system.

aik099:
> We could also update the example selenium version to 4, since this driver actually supports it.

@uuf6429 , done.

> I'm a fan of running README code in tests, to ensure that that code remains executable and up-to-date with our changes. If we can agree on this, I can create an issue to set that up later.

@uuf6429 , Not sure how you can verify, that code actually works considering that:

1. we're area using a non-existing URL in the example;
2. the actions we're doing on the page also are non-real;
3. executing example code from within a test suite isn't the same as the real user is executing it (standalone PHP file), where no classes are auto-loaded yet.

PR #16, opened 2024-03-03
Add driver factory and extension

### LINKED ISSUE #15: Connect this driver to the "Behat\MinkExtension" project
Currently, there is no way to use this driver in the Behat, because there is no corresponding driver in the https://github.com/Behat/MinkExtension/tree/master/src/Behat/MinkExtension/ServiceContainer/Driver .

The driver key would be `webdriver-classic` instead of `selenium2`. The current Selenium2 driver could be used as a starting point.

@uuf6429 , please submit a PR to that repo.

uuf6429:
Closes #15.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
         "php": ">=7.4",
         "ext-json": "*",
         "behat/mink": "^1.11@dev",
-        "php-webdriver/webdriver": "^1.14"
+        "php-webdriver/webdriver": "^1.14",
+        "behat/mink-extension": "^2.3"
     },
     "require-dev": {
         "mink/driver-testsuite": "dev-master",
@@ -24,7 +25,8 @@
         "phpstan/phpstan-phpunit": "^1.3",
         "phpunit/phpunit": "^9.6.8",
         "symfony/error-handler": "^5.4 || ^6.0",
-        "jetbrains/phpstorm-attributes": "^1.0"
+        "jetbrains/phpstorm-attributes": "^1.0",
+        "friends-of-behat/mink-extension": "^2.3"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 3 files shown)


aik099:
@uuf6429 , Adding the `behat/mink-extension` as a driver dependency assumes, that the only way to use the driver is through the Mink+Behat combo. This is incorrect.

Mink moved away from the `Behat` GitHub organization into its own `minkphp` GitHub organization a long time ago. I guess the reason for that was that Mink can be used on its own and not only through the Behat. The existence of these projects proves it:

* https://github.com/qa-tools/qa-tools/
* https://github.com/minkphp/phpunit-mink/

That might indirectly answer your question in https://github.com/minkphp/webdriver-classic-driver/issues/15#issuecomment-1963472897 as well.

aik099 (changes requested):
The Behat needs Mink, not the other way around. In this point of view the PR needs to be made into the https://github.com/FriendsOfBehat/MinkExtension as we've discussed in the associated issue.

uuf6429:
> The Behat needs Mink, not the other way around. In this point of view the PR needs to be made into the https://github.com/FriendsOfBehat/MinkExtension as we've discussed in the associated issue.

That feels like yet another problem of the `behat/mink-extension` package. As [I said earlier](https://github.com/minkphp/webdriver-classic-driver/issues/15#issuecomment-1963472897), the factories shouldn't be behat(/mink-extension)-specific.

What maybe makes sense in this case is to remove `behat/mink-extension` from `require`, leave the FriendsOfBehat one in `require-dev` and also add it to `suggests`. What do you think?

stof:
This does not belong here. This should go inside the MinkExtension repository (or in a separate extension, but as this is an official Mink driver, it can go in the main MinkExtension).

Mink itself should not depend on Behat extensions.

PR #19, opened 2024-03-05
Improve readme

uuf6429:


README.md
```diff
@@ -33,29 +66,23 @@ $session->getPage()->findLink('Chat')->click();
...
+        ```shell
+        curl -L https://github.com/SeleniumHQ/selenium/releases/download/selenium-4.18.0/selenium-server-4.18.1.jar > selenium-server-4.18.1.jar
+        java -jar selenium-server-4.18.1.jar standalone
+        ```
 2. Start WebServer by running
-    ``` bash
+    ```shell
     ./vendor/bin/mink-test-server
     ```
 3. Start PhpUnit
```
stof, line 85:
```suggestion
3. Run PHPUnit
```

README.md
```diff
@@ -1,15 +1,48 @@
 # Mink WebDriver Classic Driver
 
+[![PHP Version](http://poser.pugx.org/mink/webdriver-classic-driver/require/php)](https://packagist.org/packages/mink/webdriver-classic-driver)
```
aik099, line 3: Does this work?

README.md
```diff
@@ -33,29 +66,23 @@ $session->getPage()->findLink('Chat')->click();
...
 2. Start WebServer by running
-    ``` bash
+    ```shell
     ./vendor/bin/mink-test-server
     ```
-3. Start PhpUnit
-    ```bash
+3. Run PHPUnit
+    ```shell
     ./vendor/bin/phpunit -v --coverage-clover=coverage.clover
```
aik099, line 87: Please remove all the arguments:

* `--coverage-clover` - I doubt, that developers would require coverage to be collected on each test run;
* `-v` - not sure what it does.

aik099 (changes requested):

README.md
uuf6429: Right, how did we end up with that in the first place? 😄

README.md
uuf6429: I noticed that, it does work for other packages, e.g.

http://poser.pugx.org/phpunit/phpunit/require/php -> phpunit

My guess is that we need to make a packagist release (by adding a git(hub) tag) for it to work the first time.

aik099: I hope so.

aik099 (approved):

stof:
Isn't the "usage" section missing some information about the need to run a Selenium server ?

README.md
uuf6429: This reminds me, @stof could it make sense to tag a new pre-release?

I think the project is in a good enough state for that.

uuf6429:
> Isn't the "usage" section missing some information about the need to run a Webdriver server ?

How would you propose that would look like? It feels like it might be a repeat of the testing section _"use docker and this image OR download and run selenium, chrome-driver, chrome etc"_.

stof:
@uuf6429 people wanting to *use* the driver might not read the `Testing` section (which might even belong to a `CONTRIBUTING.md` file instead)

uuf6429:
> @uuf6429 people wanting to _use_ the driver might not read the `Testing` section (which might even belong to a `CONTRIBUTING.md` file instead)

@stof I'm not saying we shouldn't have that, I was just asking how it would look like, and/or if it's fine to copy it over from the testing section.

stof:
We could copy it from the Testing section (maybe with a sentence mentioning that you could also run the browser-specific driver directly instead of Selenium depending on your use case)

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+    1. **With Docker** - A `docker-compose.yml` file with sensible defaults is already provided , so you can just run:
+         ```shell
+         docker-compose up
+         ```
+    2. **With Java (Native)** - This would take more work, but performs better. Get started by running:
+        ```shell
+        curl -L https://github.com/SeleniumHQ/selenium/releases/download/selenium-4.18.0/selenium-server-4.18.1.jar > selenium-server-4.18.1.jar
+        java -jar selenium-server-4.18.1.jar standalone
+        ```
+2. Finally, you can simply run the tests with:
```
uuf6429, line 36: @aik099 We do not need to manually run the mink test server in this project, it is started automatically by `tests/bootstrap.php` (and terminated when testing ends), so I removed the additional point we had here before.

LICENSE
```diff
@@ -1,3 +1,5 @@
+MIT License
```
uuf6429, line 1: Added it just to be clear (we also have it mentioned in composer.json)

README.md
```diff
@@ -1,15 +1,50 @@
...
+                browser_name: chrome
+                base_url: https://example.com/
+                sessions:
+                    default:
+                        webdriver-classic: ~
+    ```
+
+## Standalone Usage Example
+
+You'll need to set up Selenium. Refer to [Contributing \ Testing](.github/CONTRIBUTING.md#testing) to see how.
```
uuf6429, line 45: @stof this isn't what we decided on earlier, but I was thinking that we should avoid huge README when possible, and in this case might as well link to the contrib file. Wdyt?

composer.json
```diff
@@ -40,5 +40,15 @@
...
         }
+    },
+    "scripts": {
+        "lint": [
+            "Composer\\Config::disableProcessTimeout",
+            "vendor/bin/phpstan"
+        ],
+        "test": [
+            "Composer\\Config::disableProcessTimeout",
+            "vendor/bin/phpunit"
```
uuf6429, line 51: There are two main reasons to do this:
1. less things to maintain - We won't have mention the tools we're using all over the place; just tell contributors to run the composer script.
2. it follows some well known or at least, common conventions, that e.g. static code analysis is typically handled by a "lint" task in the project "task manager".

.github/CONTRIBUTING.md
aik099: I wasn't aware of this. I have doubts about this approach (starting a Web Server in a test suite).

I've opened https://github.com/minkphp/webdriver-classic-driver/issues/22 issue about this.

LICENSE
aik099: Is there a specific need to add this since GitHub shows this anyway?

CHANGELOG.md
```diff
@@ -1,3 +0,0 @@
-## 1.0.0 / unreleased
-
-Initial release of the driver
```
aik099, line 3: Why you've deleted this file?

aik099 (changes requested):

README.md
stof: I don't think this makes sense as is, especially given that the contributing instructions suggest `docker compose up` which won't work as is for projects using the driver.

I suggest

> You need to run a WebDriver server, either using Selenium or the browser driver directly.

With `Selenium` being a link to the Selenium website.

README.md
```diff
@@ -1,15 +1,50 @@
...
+    ```yaml
+    # behat.yml
+    default:
+        extensions:
+            Behat\MinkExtension\Extension:
+                browser_name: chrome
+                base_url: https://example.com/
+                sessions:
+                    default:
+                        webdriver-classic: ~
```
stof, line 40: I don't think this readme should document the configuration of the Behat MinkExtension (this is something documented in the MinkExtension documentation instead).

LICENSE
uuf6429: @aik099 I'm not sure where GitHub gets that information from, I'm suspecting pattern matching the license text? 😅 

@stof this seems subjective. The heading [here is shown](https://choosealicense.com/licenses/mit/), but [not here](https://github.com/composer/composer/blob/main/LICENSE). Here are a few more: https://github.com/search?q=%22MIT+License%22+path%3ALICENSE&type=code&ref=advsearch

CHANGELOG.md
uuf6429: We should just use git(hub)'s release mechanism (which is necessary for packagist anyway), and not duplicate the funcitonality.

This is to avoid questions later like:
- is the necessary
- what's the format
- what if it's forgotten
- mentioning it in other places (like contributing.md)

README.md
uuf6429: I would normally agree, but then the `webdriver-classic` key would not be documented anywhere.

CHANGELOG.md
aik099: OK. Let's see how it works out in practice. We can always add that file later on.

README.md
uuf6429: Good point, the docker-compose file we provided is specifically for testing (and they won't actually see that in their project).

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+```shell
+composer run lint
+```
+
+## Testing
+
+1. Firstly, you will need to run Selenium and a Web Browser (and/or perhaps a driver in between):
+    1. **With Docker** - A `docker-compose.yml` file with sensible defaults is already provided , so you can just run:
+         ```shell
+         docker-compose up
```
stof, line 29: I suggest using `docker composer up`, which is the modern way (compose as a plugin of the docker CLI instead of a separate executable)

README.md
aik099: Do you know how this problem is solved in the MinkSelenium2Driver? It's `selenium2` key should be advertised somewhere.

aik099 (changes requested):

README.md
uuf6429: There's one here: https://docs.behat.org/en/v2.5/cookbook/behat_and_mink.html#:~:text=The%20easiest%20way%20to%20get%20started%20is%20to%20go%20with%20goutte%20and%20selenium2%20drivers%2C%20but%20note%20that%20there%E2%80%99s%20bunch%20of%20other%20drivers%20available%20for%20Mink%20%2D%20read%20about%20them%20in%20Mink%20documentation.

There isn't any mention in the mink documentation as far as I could see: https://mink.behat.org/en/latest/drivers/selenium2.html

@stof is also right - that key is not even in this repository; it's actually defined in `FriendsOfBehat/MinkExtension`.

I'd like to keep it out of this repository/readme, but I have no guarantees it will be documented elsewhere.

uuf6429: Removed the change.

README.md
```diff
@@ -1,15 +1,39 @@
...
+composer require "behat/mink" "mink/webdriver-classic-driver"
+```
+
+### For Behat/BDD Usage
+
+1. Add all the relevant packages - typically this will be for testing your application, so you'd want to use `--dev`.
+    ```shell
+    composer require --dev "behat/behat" "behat/mink" "friends-of-behat/mink-extension" "mink/webdriver-classic-driver"
+    ```
+2. Configure Behat to enable the Mink Extension and set it to use this drive.
```
stof, line 30:
```suggestion
2. Configure Behat to enable the Mink Extension and set it to use this driver.
```

stof (approved):

aik099:
@uuf6429 , could you please remind me where the sample `.behat.yml` was supposed to live?

uuf6429:
> @uuf6429 , could you please remind me where the sample `.behat.yml` was supposed to live?

Didn't we say we're not providing a sample behat config? Otherwise I don't know what's this about..it's been quite a while now.

aik099:
> > @uuf6429 , could you please remind me where the sample `.behat.yml` was supposed to live?
> 
> @aik099 Didn't we say we're not providing a sample behat config? Otherwise I don't know what's this about..it's been quite a while now.

Would that sample configuration file be located in MinkExtension repo?

composer.json
stof: I'm still not convinced because composer is quite bad as a task executor. It is not really meant to be a generic task executor

uuf6429: @stof could you back up that claim? It might not be _ideal_, but as far as I know, it works quite well.

Besides, anyone could copy those commands and run them in their console, if they really want to. 
It's still better than duplicating this all over the place (github action: `php vendor/bin/phpunit yadda yadda`, contributing.md: "Run `php vendor/bin .....` to run tests...if you're on windows run ....." - repeated for each tool)

uuf6429:
> Would that sample configuration file be located in MinkExtension repo?

@aik099 Feel more like it should be in the documentation somewhere. If we did put it in the repo, it would probably be lost (export-ignore'd), so devs wouldn't find it anyway.

stof:
MinkExtension has a `doc` folder, where it documents all drivers it supports with its factories. That's where things should go.
The Mink driver themselves should not care about the Behat extension that is using them.

uuf6429:
> MinkExtension has a `doc` folder, where it documents all drivers it supports with its factories. That's where things should go. The Mink driver themselves should not care about the Behat extension that is using them.

It's the first time I see that - totally wasn't aware of it. The last change there is 7 years ago, by the way (that said it looks very useful, just maybe not very visible).

README.md
```diff
@@ -1,15 +1,39 @@
...
+composer require "behat/mink" "mink/webdriver-classic-driver"
+```
+
+### For Behat/BDD Usage
+
+1. Add all the relevant packages - typically this will be for testing your application, so you'd want to use `--dev`.
+    ```shell
+    composer require --dev "behat/behat" "behat/mink" "friends-of-behat/mink-extension" "mink/webdriver-classic-driver"
+    ```
+2. Configure Behat to enable the Mink Extension and set it to use this driver.
```
aik099, line 30: Would posting a link on https://github.com/FriendsOfBehat/MinkExtension/blob/master/doc/index.rst page, that will explain how to connect this driver specifically to Behat hurt?

// cc @stof .

uuf6429: I think it's a good idea.

PR #23, opened 2024-03-14
Require symfony/process as dev dependency

uuf6429:


oleg-andreyev:
Reason?

uuf6429:
@oleg-andreyev it's [used here](https://github.com/minkphp/webdriver-classic-driver/blob/main/tests/bootstrap.php), as said in the "mentioned" link above. The version update is just to avoid restrictions for nothing.

composer.json
```diff
@@ -23,7 +23,8 @@
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-phpunit": "^1.3",
         "phpunit/phpunit": "^9.6.8",
-        "symfony/error-handler": "^5.4 || ^6.0",
+        "symfony/error-handler": "^5.4 || ^6.0 || ^7.0",
+        "symfony/process": "^5.4 || ^6.0 || ^7.0",
```
aik099, line 27: If this driver was already released, then I would suggest moving the `|| ^7.0` part to a different PR for adding Symfony 7 support.

Except for that, everything looks good.

oleg-andreyev (approved):

stof (approved):

PR #24, opened 2024-03-21
Don't iterate windows, when already at correct window

aik099:
I've noticed, that when trying to resize/maximize the current window several calls are made to the Selenium server to verify if we're at the right window. IMO they're unnecessary because we're surely at the correct window already.

P.S.
There is no test to verify, that resizing/maximizing non-current window works.


uuf6429 (approved):
I think you're right.

stof (approved):

PR #27, opened 2024-04-07
Harden window closing code upon session reset

### LINKED ISSUE #25: The `Session::reset` closes main window, when window was renamed
How to reproduce:

1. start a session either by calling the `Session::visit`
   * an empty window is opened with an empty `window.name` (because no URL is opened yet)
   * initial window handle becomes `w3cwh:{selenium_window_handle_here}`
   * the page URL is visited
   * the page sets it's `window.name` to the desired value
2. open some other popups from that page
3. call the `Session::reset`:
   * the `Session::swithToWindow()` works correctly, because it uses selenium-handle-based window identifier
   * code, that cycles through the opened windows isn't identifying an initial window (because it has a different name) and such and is closing it

For now the only option, that might work is to consider 1st opened window as initial (the 1st window handle returned) and never close it.

Related issue: #26

### LINKED ISSUE #26: The `Session::reset` throws an exception when main window has name
When the main window has a name and a couple of other windows are open, then the `Session::reset` call throws a `NoSuchWindow` exception, because it tries to determine the main window handle by its name after a popup window was just deleted while Selenium is in context of that deleted window (Web View).

Maybe for the internal window manipulations, the driver should use only window handles.

Related issue: #25

P.S.
I stumbled upon this error after fixing the #25 problem locally.

aik099:
Problems solved:

* initial window is closed upon session reset, when it's name was changed through the JavaScript after opening (Closes #25)
* an `NoSuchWindow` exception being thrown upon session reset, when initial window name was changed through the JavaScript after opening (Closes #26)

Test failure is unrelated.


tests/Custom/SessionResetTest.php
```diff
@@ -0,0 +1,62 @@
...
+
+        $expectedInitialWindowName = $session->evaluateScript('window.name');
+
+        $windowNames = $session->getWindowNames();
+        $this->assertCount(3, $windowNames);
+
+        $session->reset();
+
+        $windowNames = $session->getWindowNames();
+        $this->assertCount(1, $windowNames);
```
uuf6429, line 34: I'm not fully sure I understand the problem being solved here.

Why are you expecting 1 window after a session reset?

aik099: Because the `WebdriverClassicDriver::reset` method closes all opened windows, except a window, that opened upon session start.

I haven't changed this behavior (I like that). I've added a test to ensure it's working as expected.

uuf6429: OK, so to be clear, the correct behaviour is to keep the initial window open?

Should the window refresh/reload? Otherwise it's possible it "contaminates" the session after reset (e.g. if the window contains some web app that saves session data as cookies/local-storage periodically). 🤔

src/WebdriverClassicDriver.php
```diff
@@ -136,16 +136,22 @@ public function stop(): void
      */
     public function reset(): void
     {
+        $webDriver = $this->getWebDriver();
+
         // switch to default window..
-        $this->switchToWindow();
+        $initialWindow = substr((string)$this->initialWindowName, strlen(self::W3C_WINDOW_HANDLE_PREFIX));
```
uuf6429, line 142: This will still not work because `initialWindowName` may contain a name and not a handle, since `$this->getWindowName()` can return a window name (instead of handle). The substr would end up slicing out a part of the name.

tests/Custom/SessionResetTest.php
```diff
@@ -0,0 +1,62 @@
...
+namespace Mink\WebdriverClassDriver\Tests\Custom;
+
+use Behat\Mink\Tests\Driver\TestCase;
+
+class SessionResetTest extends TestCase
+{
+    /**
+     * @dataProvider initialWindowNameDataProvider
+     */
+    public function testSessionResetClosesWindows(?string $initialWindowName): void
```
uuf6429, line 12: @stof Would it make sense to put this test in driver-testsuite?

src/WebdriverClassicDriver.php
uuf6429: Here's a patch, it fixes several things:
```patch
Subject: [PATCH] use window handles internally
---
Index: src/WebdriverClassicDriver.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/WebdriverClassicDriver.php b/src/WebdriverClassicDriver.php
--- a/src/WebdriverClassicDriver.php	(revision 71c242b36e92fe0fe3891b2700a41eb629b61550)
+++ b/src/WebdriverClassicDriver.php	(date 1712604857594)
@@ -78,7 +78,7 @@
 
     private string $webDriverHost;
 
-    private ?string $initialWindowName = null;
+    private ?string $initialWindowHandle = null;
 
     /**
      * @param string $browserName One of 'edge', 'firefox', 'chrome' or any one of {@see WebDriverBrowserType} constants.
@@ -104,7 +104,7 @@
         try {
             $this->webDriver = RemoteWebDriver::create($this->webDriverHost, $this->desiredCapabilities);
             $this->applyTimeouts();
-            $this->initialWindowName = $this->getWindowName();
+            $this->initialWindowHandle = $this->webDriver->getWindowHandle();
         } catch (\Throwable $e) {
             throw new DriverException("Could not start driver: {$e->getMessage()}", 0, $e);
         }
@@ -136,13 +136,17 @@
      */
     public function reset(): void
     {
-        // switch to default window..
-        $this->switchToWindow();
-        // ..and close all other windows
-        foreach ($this->getWindowNames() as $name) {
-            if ($name !== $this->initialWindowName) {
-                $this->withWindow($name, fn() => $this->getWebDriver()->close());
-            }
+        $webDriver = $this->getWebDriver();
+
+        // close all windows except the initial one
+        foreach ($webDriver->getWindowHandles() as $windowHandle) {
+            if ($windowHandle === $this->initialWindowHandle) {
+                continue;
+            }
+
+            $webDriver->switchTo()->window($windowHandle);
+            $webDriver->close();
+            $this->switchToWindow();
         }
 
         $this->getWebDriver()->manage()->deleteAllCookies();
@@ -175,13 +179,9 @@
 
     public function switchToWindow(?string $name = null): void
     {
-        if ($name === null) {
-            $name = $this->initialWindowName;
-        }
-
-        if (is_string($name)) {
-            $name = $this->getWindowHandleFromName($name);
-        }
+        $name = $name === null
+            ? $this->initialWindowHandle
+            : $this->getWindowHandleFromName($name);
 
         $this->getWebDriver()->switchTo()->window((string)$name);
     }
```

tests/Custom/SessionResetTest.php
aik099: > OK, so to be clear, the correct behaviour is to keep the initial window open?

Yes, that is how you implemented it. Do you remember this? There was no similar code in the MinkSelenim2Driver.

> Should the window refresh/reload? Otherwise it's possible it "contaminates" the session after reset (e.g. if the window contains some web app that saves session data as cookies/local-storage periodically). 🤔

The `reset` method clears cookies and closes any windows opened by the test. It's expected, that the next test will visit the desired page and since the cookies are absent everything will reset by itself (this happens in the driver test suite).

src/WebdriverClassicDriver.php
aik099: > This will still not work because `initialWindowName` may contain a name and not a handle, since `$this->getWindowName()` can return a window name (instead of handle). The substr would end up slicing out a part of the name.

When the initial window name is populated (after the session was just started with an empty window and no URL visited yet) the window doesn't have a name. It even doesn't have any HTML in it. So I'm pretty certain that `substr` would work.

> Here's a patch, it fixes several things:

I like it. I've made several adjustments to:

* account for code changes since the patch creation time;
* keep initial `$this->switchToWindow();` call, because it would save the day when the test finishes at the point when the currently active window is removed and WebDriver would throw the `NoSuchWindow` exception on any command we perform.

tests/Custom/SessionResetTest.php
aik099: I've added this test into the driver's test suite instead of the Mink driver test suite because the window closing behavior upon session reset is non-standard or advertised anywhere.

tests/Custom/SessionResetTest.php
uuf6429: > Do you remember this?

Nope. 😅 I do remember that the windows had to be closed at some point, but I really thought that was the same behaviour across all drivers.

> It's expected, that the next test will visit the desired page

Ok, sounds reasonable. 👍

src/WebdriverClassicDriver.php
uuf6429: > keep initial $this->switchToWindow(); call, because it would save the day when the test finishes at the point when the currently active window is removed and WebDriver would throw the NoSuchWindow exception on any command we perform.

Aah, I see. If there's only one window, then the foreach loop would skip switching to it and might stay stuck on a closed window. Maybe this would be a cleaner solution (and perhaps also better performing)?
```php
foreach ($webDriver->getWindowHandles() as $windowHandle) {
    if ($windowHandle !== $this->initialWindowHandle) {
        continue;
    }
    $webDriver->switchTo()->window($windowHandle);
    $webDriver->close();
}
$this->switchToWindow();
```

aik099: I've played around a bit and surprisingly:

* the `getWindowHandle` call made inside a closed window context fails with a `NoSuchWindow` exception;
* the `getWindowHandles` call made inside a closed window context succeeds;
* attempts to switch windows from inside a closed window context succeeds.

This proves, that calling `$this->switchToWindow();` before/inside the loop isn't necessary.

I've updated the PR code to include the proposed changes.

src/WebdriverClassicDriver.php
```diff
@@ -175,13 +182,9 @@ public function back(): void
...
     public function switchToWindow(?string $name = null): void
     {
-        if ($name === null) {
-            $name = $this->initialWindowName;
-        }
-
-        if (is_string($name)) {
-            $name = $this->getWindowHandleFromName($name);
-        }
+        $name = $name === null
```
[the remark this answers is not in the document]
aik099: Done.

stof (approved):

aik099:
Thank you, @stof .

PR #29, opened 2024-08-05
Use chromium that run ARM and x86_64, and use dev webdriver.

saidatom:


docker-compose.yml
```diff
@@ -1,8 +1,6 @@
-version: "3"
-
 services:
   selenium:
-    image: ${SELENIUM_IMAGE:-selenium/standalone-chrome:4}
+    image: ${SELENIUM_IMAGE:-selenium/standalone-chromium:4}
```
uuf6429, line 3: The main reason `chrome` was used by default is mainly the [high market share](https://gs.statcounter.com/browser-market-share) which also means many users would expect such a browser by default (we had `firefox` before). That said, it has been defined as a variable so the user can always use a different image entirely (even a non-selenium/standalone image). Therefore I would recommend removing this change.

uuf6429 (changes requested):
Thanks for the PR! I've left a few review comments..

oleg-andreyev (approved):

uuf6429 (approved):
Pretty sure the failed tests are unrelated to these changes, so the PR looks fine IMO.
I'll look at the failures separately.

PR #32, opened 2024-11-01
Use innerText instead of selenium call

### LINKED ISSUE #30: getText doesn't seem to always works
While trying to track a weird problem, it seemed stem from `getText()` returning an empty string, even when I can see the correct element is being selected and the expected content is show when I use `getOuterHtml()` or `getHtml()`.

There isa major difference in the implementation between those:
https://github.com/minkphp/webdriver-classic-driver/blob/9567c6e77c2b89dcdc5a5f567ce826c1a831de1f/src/WebdriverClassicDriver.php#L297
vs:
https://github.com/minkphp/webdriver-classic-driver/blob/9567c6e77c2b89dcdc5a5f567ce826c1a831de1f/src/WebdriverClassicDriver.php#L304

I was able to solve my problem by using `return $this->getElementDomProperty($this->findElement($xpath), 'textContent');`.

I have a feeling the original implementation was the same in all three cases, so now I'm wondering which case is the really correct one, when the dedicated cases seems to misbehave.

uuf6429:
Fixes #30

### DIFF
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -294,14 +294,14 @@ class WebdriverClassicDriver extends CoreDriver
         #[Language('XPath')]
         string $xpath
     ): string {
-        return str_replace(["\r", "\n"], ' ', $this->findElement($xpath)->getText());
+        return $this->getElementDomProperty($this->findElement($xpath), 'innerText');
     }
 
     public function getHtml(
         #[Language('XPath')]
         string $xpath
     ): string {
-        return $this->getElementDomProperty($this->findElement($xpath), 'innerText');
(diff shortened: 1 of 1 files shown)


uuf6429:
@aik099 the remaining failures were already happening before. I made a new issue for that: #33

aik099 (approved):
Looks good.

aik099:
You might be interested in porting https://github.com/minkphp/MinkSelenium2Driver/pull/328 fix to this driver (in this or separate PR).

uuf6429:
Done

aik099:
Merging. Thanks @uuf6429 .

PR #34, opened 2024-11-01
Use port from host var

uuf6429:


tests/bootstrap.php
```diff
@@ -4,10 +4,14 @@
...
 
+$minkTestServerPort = isset($_SERVER['WEB_FIXTURES_HOST'])
+    ? parse_url($_SERVER['WEB_FIXTURES_HOST'], PHP_URL_PORT)
+    : '8002';
+
 $minkTestServer = new Process([
     PHP_BINARY,
     '-S',
-    '0.0.0.0:8002',
+    "0.0.0.0:$minkTestServerPort",
```
aik099, line 14: It works, but for performance reasons maybe it's better to be rewritten without double quote usage as:

```php
'0.0.0.0:' . $minkTestServerPort,
```

aik099 (changes requested):

tests/bootstrap.php
uuf6429: It should be a well know fact that string interpolation in PHP has no performance impact and anyway that code executes once (it's in the bootstrap).

Anyway, I applied the change.

aik099: Thank you.

aik099 (approved):

Issue #10, opened 2024-02-22
Improve test coverage


aik099:
Browsing through the driver code I've noticed, that most of the error reporting isn't covered by tests. Without this, we can't be sure if the code actually works.

Either the https://github.com/minkphp/driver-testsuite needs to be improved or this repo custom tests.

stof:
I think this might be a mix of both.

The error handling for `start` and `stop` should be covered in custom tests (as the failure modes there are specific to each driver).
The error handling of `getWebdriver` should as well because the DriverInterface makes it an undefined behavior to call methods on a non-started driver. It does not require getting a DriverException in such case.

Other cases might correspond to holes in the coverage of invalid usages in the shared testsuite (for instance, we added testing for some invalid form field value types among the latest improvements of that testsuite). Those should be implemented in the shared testsuite to improve coverage in all drivers.

PR #37, opened 2024-11-02
Improve start/stop test coverage

uuf6429:
Relates to #10.

### DIFF
--- a/tests/Custom/WebDriverTest.php
+++ b/tests/Custom/WebDriverTest.php
@@ -2,7 +2,6 @@
 
 namespace Mink\WebdriverClassicDriver\Tests\Custom;
 
-use Behat\Mink\Driver\DriverInterface;
 use Behat\Mink\Exception\DriverException;
 use Mink\WebdriverClassicDriver\Tests\WebdriverClassicConfig;
 use Mink\WebdriverClassicDriver\WebdriverClassicDriver;
@@ -10,7 +9,7 @@ use PHPUnit\Framework\TestCase;
 
 class WebDriverTest extends TestCase
 {
-    private DriverInterface $driver;
+    private WebdriverClassicDriver $driver;
 
     protected function setUp(): void
     {
@@ -88,10 +87,6 @@ class WebDriverTest extends TestCase
 
     public function testClassicDriverCanProvideBrowserName(): void
     {
-        if (!$this->driver instanceof WebdriverClassicDriver) {
-            $this->markTestSkipped('This test applies to WebdriverClassicDriver only');
-        }
-
         $this->assertSame(
             WebdriverClassicConfig::getInstance()->getBrowserName(),
             $this->driver->getBrowserName()
--- a/tests/Custom/WebDriverTest.php
+++ b/tests/Custom/WebDriverTest.php
@@ -30,7 +30,7 @@ class WebDriverTest extends TestCase
     public function testDriverMustBeStartedBeforeUse(): void
     {
         $this->expectException(DriverException::class);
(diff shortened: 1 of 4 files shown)


tests/Custom/WebDriverTest.php
```diff
@@ -2,32 +2,94 @@
...
+
+        $this->expectException(DriverException::class);
+        $this->expectExceptionMessage('Could not start driver: An upstream error');
+
+        $driver->start();
+    }
+
+    public function testDriverCatchesUpstreamErrorsDuringStop(): void
+    {
+        $driver = $this->createPartialMock(WebdriverClassicDriver::class, ['createWebDriver', 'isStarted', 'getWebDriver']);
```
aik099, line 72: Partial mocks usually indicate bad architectural decisions (that's why Prophecy library disallows them).

If we would be using some kind of factory, that internally called the `RemoteWebDriver::create` method in the `\Mink\WebdriverClassicDriver\WebdriverClassicDriver::start`, then we could have injected a new factory for testing, that returned a mock of the `RemoteWebDriver` class.

Maybe there are other architectural solutions to this.

tests/WebdriverClassicConfig.php
```diff
@@ -18,12 +17,16 @@ public static function getInstance(): self
...
+    public function createDriver(): WebdriverClassicDriver
     {
-        $browser = getenv('WEB_FIXTURES_BROWSER') ?: WebdriverClassicDriver::DEFAULT_BROWSER;
         $seleniumHost = $_SERVER['DRIVER_URL'];
 
-        return new WebdriverClassicDriver($browser, [], $seleniumHost);
+        return new WebdriverClassicDriver($this->getBrowserName(), [], $seleniumHost);
+    }
+
+    public function getBrowserName(): string
```
aik099, line 27: Why this method was added?

aik099 (changes requested):

tests/Custom/WebDriverTest.php
uuf6429: In this case it's fine, having a factory for the base driver feels like overengineering to me.

tests/WebdriverClassicConfig.php
uuf6429: There's a test for the driver's `getBrowserName()` method which uses this.

It's needed in case we run the tests for other browsers (which we do already). I could replicate the function's body in the tests, but it's pointless duplication.

tests/Custom/WebDriverTest.php
aik099: OK.

tests/WebdriverClassicConfig.php
aik099: The `$this->getSession()->getDriver()->getBrowserName()` code can be used instead of creating `WebdriverClassicConfig::getBrowserName` method.

aik099 (changes requested):

tests/WebdriverClassicConfig.php
uuf6429: It's used for testing that method (`getDriver()->getBrowserName()`).

Using `assertSame(getDriver()->getBrowserName(), getDriver()->getBrowserName())` wouldn't make a lot of sense. 😆

That method is needed because this would not work when tests are running against e.g. firefox:
`assertSame('chrome', getDriver()->getBrowserName())`

aik099: Ups.

I understand now what is being tested. The code is OK.

aik099 (approved):

PR #38, opened 2024-11-02
Support MinkSelenium2Driver page timeout

### LINKED ISSUE #36: Support for `page load` and `pageLoad` timeout type
Currently, the `WebdriverClassicDriver::applyTimeouts` method supports these timeout types:
* `script`;
* `implicit`;
* `page`.

The `script` and `implicit` are timeout types shared by Selenium Server 2 and W3C-compatible Selenium Server (3 and 4).

However, Selenium 2 uses `page load` instead of page and W3C-compatible Selenium Server (3 and 4) uses `pageLoad`.

Proposing to add support for `page load` and `pageLoad` values to ease migration from MinkSelenium2Driver, which supplied given timeout types as-is to the Selenium Server.

P.S.
* It's better to remove `page` timeout to avoid confusion, but that might look like a BC break.
* FYI: The `script` timeout type is completely useless because it only affects `\Facebook\WebDriver\Remote\RemoteWebDriver::executeAsyncScript` method usages, that aren't found in this driver.

uuf6429:
Fixes #36.

### DIFF
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -981,7 +981,7 @@ class WebdriverClassicDriver extends CoreDriver
             throw new DriverException("Error setting timeout: {$e->getMessage()}", 0, $e);
         }
 
-        $timeouts = $this->webDriver->executeCustomCommand(
+        $timeouts = $this->getWebDriver()->executeCustomCommand(
             '/session/:sessionId/timeouts',
             'GET'
         );
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -980,15 +980,6 @@ class WebdriverClassicDriver extends CoreDriver
         } catch (\Throwable $e) {
             throw new DriverException("Error setting timeout: {$e->getMessage()}", 0, $e);
         }
-
-        echo 'Session ID: ' . $this->getWebDriverSessionId() . PHP_EOL;
-
-        $timeouts = $this->getWebDriver()->executeCustomCommand(
-            '/session/:sessionId/timeouts',
-            'GET'
-        );
-
-        print_r($timeouts);
     }
 
     /**
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -77,7 +77,16 @@ class WebdriverClassicDriver extends CoreDriver
 
     private DesiredCapabilities $desiredCapabilities;
 
-    private array $timeouts = [];
+    /**
+     * Default timeouts as per [W3C Specification](https://w3c.github.io/webdriver/#timeouts), in milliseconds.
+     *
+     * @var array<string, null|int>
+     */
+    private array $timeouts = [
+        'script' => 30000,
+        'page' => 300000,
+        'implicit' => 0,
+    ];
 
     private string $webDriverHost;
 
@@ -967,7 +976,7 @@ class WebdriverClassicDriver extends CoreDriver
                             "Using \"$type\" timeout type is deprecated, please use \"page\" instead",
                             E_USER_DEPRECATED
                         );
-                        // no break
+                    // no break
 
                     case 'page':
                         $timeouts->pageLoadTimeout($param / 1000);
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -967,8 +967,7 @@ class WebdriverClassicDriver extends CoreDriver
                             "Using \"$type\" timeout type is deprecated, please use \"page\" instead",
                             E_USER_DEPRECATED
                         );
-                        $timeouts->pageLoadTimeout($param / 1000);
-                        break;
+                        // no break
 
                     case 'page':
                         $timeouts->pageLoadTimeout($param / 1000);
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -980,6 +980,13 @@ class WebdriverClassicDriver extends CoreDriver
         } catch (\Throwable $e) {
             throw new DriverException("Error setting timeout: {$e->getMessage()}", 0, $e);
         }
+
+        $timeouts = $this->webDriver->executeCustomCommand(
+            '/session/:sessionId/timeouts',
+            'GET'
+        );
+
+        print_r($timeouts);
     }
 
     /**
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -77,16 +77,7 @@ class WebdriverClassicDriver extends CoreDriver
 
     private DesiredCapabilities $desiredCapabilities;
 
-    /**
-     * Default timeouts as per [W3C Specification](https://w3c.github.io/webdriver/#timeouts), in milliseconds.
-     *
-     * @var array<string, null|int>
-     */
-    private array $timeouts = [
-        'script' => 30000,
-        'page' => 300000,
-        'implicit' => 0,
-    ];
+    private array $timeouts = [];
 
     private string $webDriverHost;
 
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -981,6 +981,8 @@ class WebdriverClassicDriver extends CoreDriver
             throw new DriverException("Error setting timeout: {$e->getMessage()}", 0, $e);
         }
 
+        echo 'Session ID: ' . $this->getWebDriverSessionId() . PHP_EOL;
+
         $timeouts = $this->getWebDriver()->executeCustomCommand(
             '/session/:sessionId/timeouts',
             'GET'
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -961,6 +961,9 @@ class WebdriverClassicDriver extends CoreDriver
                         $timeouts->implicitlyWait($param / 1000);
                         break;
 
+                    case 'page load':
+                    case 'pageLoad':
+                        // ☝️ For MinkSelenium2Driver backward compatibility
                     case 'page':
                         $timeouts->pageLoadTimeout($param / 1000);
                         break;
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -963,7 +963,13 @@ class WebdriverClassicDriver extends CoreDriver
 
                     case 'page load':
                     case 'pageLoad':
-                        // ☝️ For MinkSelenium2Driver backward compatibility
+                        @trigger_error(
+                            "Using \"$type\" timeout type is deprecated, please use \"page\" instead",
+                            E_USER_DEPRECATED
+                        );
+                        $timeouts->pageLoadTimeout($param / 1000);
+                        break;
+
                     case 'page':
                         $timeouts->pageLoadTimeout($param / 1000);
                         break;
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -57,8 +57,8 @@ class TimeoutTest extends TestCase
     {
         $this->driver->start();
 
-//        $this->expectDeprecation('Using "' . $type . '" timeout type is deprecated, please use "page" instead');
-        $this->driver->setTimeouts([$type => 500]);
+        $this->expectDeprecation('Using "' . $type . '" timeout type is deprecated, please use "page" instead');
+//        $this->driver->setTimeouts([$type => 500]);
 
         $this->expectException(DriverException::class);
         $this->expectExceptionMessage('Page failed to load: ');
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -57,7 +57,7 @@ class TimeoutTest extends TestCase
     {
         $this->driver->start();
 
-        $this->expectDeprecation('Using "' . $type . '" timeout type is deprecated, please use "page" instead');
+//        $this->expectDeprecation('Using "' . $type . '" timeout type is deprecated, please use "page" instead');
         $this->driver->setTimeouts([$type => 500]);
 
         $this->expectException(DriverException::class);
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -8,11 +8,7 @@ class TimeoutTest extends TestCase
 {
     protected function tearDown(): void
     {
-        $this->driver->setTimeouts([
-            'script' => 30000,
-            'page' => 300000,
-            'implicit' => 0,
-        ]);
+        $this->driver->setTimeouts([]);
 
         parent::tearDown();
     }
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -18,9 +18,9 @@ class TimeoutTest extends TestCase
 
     public function testShortTimeoutDoesNotWaitForElementToAppear(): void
     {
-        $this->driver->start();
-        $this->fail('URL is: ' . $this->driver->visit($this->pathTo('/js_test.html')));
+        $this->fail('URL is: ' . $this->pathTo('/js_test.html'));
 
+        $this->driver->start();
         $this->driver->setTimeouts(['implicit' => 0]);
 
         $this->driver->visit($this->pathTo('/js_test.html'));
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -18,8 +18,6 @@ class TimeoutTest extends TestCase
 
     public function testShortTimeoutDoesNotWaitForElementToAppear(): void
     {
-        $this->fail('URL is: ' . $this->pathTo('/js_test.html'));
-
         $this->driver->start();
         $this->driver->setTimeouts(['implicit' => 0]);
 
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -18,6 +18,8 @@ class TimeoutTest extends TestCase
 
     public function testShortTimeoutDoesNotWaitForElementToAppear(): void
     {
+        $this->fail('URL is: ' . $this->driver->visit($this->pathTo('/js_test.html')));
+
         $this->driver->start();
         $this->driver->setTimeouts(['implicit' => 0]);
 
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -6,7 +6,7 @@ use Behat\Mink\Exception\DriverException;
 
 class TimeoutTest extends TestCase
 {
-     public function testInvalidTimeoutSettingThrowsException(): void
+    public function testInvalidTimeoutSettingThrowsException(): void
     {
         $this->driver->start();
 
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -65,11 +65,9 @@ class TimeoutTest extends TestCase
         $this->driver->visit($this->pathTo('/page_load.php?sleep=2'));
     }
 
-    public static function deprecatedPageLoadDataProvider(): array
+    public static function deprecatedPageLoadDataProvider(): iterable
     {
-        return [
-            'selenium 3 style' => ['type' => 'pageLoad'],
-            'selenium 2 style' => ['type' => 'page load'],
-        ];
+        yield 'selenium 3 style' => ['type' => 'pageLoad'];
+        yield 'selenium 2 style' => ['type' => 'page load'];
     }
 }
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -59,4 +59,27 @@ class TimeoutTest extends TestCase
 
         $this->driver->visit($this->pathTo('/page_load.php?sleep=2'));
     }
+
+    /**
+     * @group legacy
+     * @dataProvider deprecatedPageLoadDataProvider
+     */
+    public function testDeprecatedShortPageLoadTimeoutThrowsException(string $type): void
+    {
+        $this->driver->start();
+
+        $this->expectDeprecation('Using "' . $type . '" timeout type is deprecated, please use "page" instead');
+        $this->driver->setTimeouts([$type => 500]);
+
+        $this->expectException(DriverException::class);
+        $this->expectExceptionMessage('Page failed to load: ');
+
+        $this->driver->visit($this->pathTo('/page_load.php?sleep=2'));
+    }
+
+    public static function deprecatedPageLoadDataProvider(): iterable
+    {
+        yield ['type' => 'pageLoad'];
+        yield ['type' => 'page load'];
+    }
 }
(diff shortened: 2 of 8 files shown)


src/WebdriverClassicDriver.php
```diff
@@ -940,6 +940,9 @@ private function applyTimeouts(): void
                     case 'implicit':
                         $timeouts->implicitlyWait($param / 1000);
                         break;
+                    case 'page load':
+                    case 'pageLoad':
+                        // ☝️ For MinkSelenium2Driver backward compatibility
```
aik099, line 945: Was an emoji added to the comment on purpose?

uuf6429: Yeah... thought it would be some harmless fun.

uuf6429: Although now that I think about it, perhaps it's more functionally useful to replace that comment with:
```php
@trigger_error(
    "Using `$type` timeout type is deprecated, please use `page` instead.",
    E_USER_DEPRECATED
);
```
What do you think?

aik099: That makes sense.

aik099 (changes requested):
Please add a deprecation notice as you suggested.

src/WebdriverClassicDriver.php
```diff
@@ -937,15 +937,24 @@ private function applyTimeouts(): void
...
+
                     case 'page load':
                     case 'pageLoad':
-                        // ☝️ For MinkSelenium2Driver backward compatibility
+                        @trigger_error(
+                            "Using \"$type\" timeout type is deprecated, please use \"page\" instead",
+                            E_USER_DEPRECATED
+                        );
+                        $timeouts->pageLoadTimeout($param / 1000);
+                        break;
```
aik099, line 952: This part isn't needed.

src/WebdriverClassicDriver.php
```diff
@@ -937,12 +937,24 @@ private function applyTimeouts(): void
...
                     case 'implicit':
                         $timeouts->implicitlyWait($param / 1000);
                         break;
+
+                    case 'page load':
+                    case 'pageLoad':
+                        @trigger_error(
+                            "Using \"$type\" timeout type is deprecated, please use \"page\" instead",
+                            E_USER_DEPRECATED
+                        );
```
aik099, line 950: Where are the tests for this code?

aik099 (changes requested):

src/WebdriverClassicDriver.php
uuf6429: I know, but I get IDE warnings that it might be a mistake ("that I forgot to break;"). 🙈

aik099: Replace that code (highlighted by this comment) with `// no break` line. It should work in PhpStorm (if you're using it) according to the https://youtrack.jetbrains.com/issue/WI-32123/Usage-of-PSR-2-comment-suggestion-on-no-break-switch-cases .

src/WebdriverClassicDriver.php
uuf6429: I did give it a try and gave up after finding that PHPUnit has deprecated...catching deprecations. 😅
https://phpunit.de/announcements/phpunit-10.html#:~:text=Test%20Outcome%20versus%20Test%20Issues

Any idea how to continue further? The only way I see is by registering our own `set_error_handler` 🙄. Although I remember there was something custom in the driver testsuite about this and symfony. 🤔

aik099: Next steps:

1. merge the #39 to `main` branch
2. rebase this PR upon the `main` branch
3. adjust the `testShortPageLoadTimeoutThrowsException` test (was added in #39) to support data provider (as was done in https://github.com/minkphp/MinkSelenium2Driver/pull/400) and when `$timeoutType` would be `page load` or `pageLoad` we'll see how current error handler (including Symfony's version of it) will work on PHPUnit 9 and PHPUnit 10

aik099: I've rebased this PR on top of `main` branch after #39 was merged. Now you can experiment with `testShortPageLoadTimeoutThrowsException`.

tests/Custom/TimeoutTest.php
```diff
@@ -76,6 +79,32 @@ public function testShortPageLoadTimeoutThrowsException(): void
 
         $this->expectException(DriverException::class);
         $this->expectExceptionMessage('Page failed to load: ');
+
         $session->visit($this->pathTo('/page_load.php?sleep=2'));
     }
+
+    /**
+     * @group legacy
```
uuf6429, line 87: The symfony deprecations helper requires the test to be in the "legacy" group, so I could not just alter the existing test.

aik099: OK.

tests/Custom/TimeoutTest.php
```diff
@@ -76,6 +79,32 @@ public function testShortPageLoadTimeoutThrowsException(): void
...
+     */
+    public function testDeprecatedShortPageLoadTimeoutThrowsException(string $type): void
+    {
+        $session = $this->getSession();
+        $driver = $session->getDriver();
+        \assert($driver instanceof WebdriverClassicDriver);
+
+        $driver->setTimeouts([$type => 500]);
+
+        $this->expectDeprecation('Using "' . $type . '" timeout type is deprecated, please use "page" instead');
```
aik099, line 98: 1. Have you checked that absense of deprecation notice causes a test to fail on both PHPUni 9.x and PHPUnit 10.x?
2. Do we test on PHPUnit 10.x at all?

aik099 (changes requested):

tests/Custom/TimeoutTest.php
uuf6429: 1. Of course...what would be the point of "expectDeprecation" otherwise? 😄 
2. Apparently not...  dependency is at `^9.6.8`. We should probably upgrade since that version is [on life support](https://phpunit.de/supported-versions.html).

Why do you ask?

Note that this is using the symfony functionality, not the original PHPUnit one (by using a trait).

uuf6429: Side note, apparently there's a lot of failures now... could it be related to a recent driver-testsuite update?
https://github.com/minkphp/webdriver-classic-driver/actions/runs/11650940031

It doesn't happen to me locally, even after updating composer dependencies. 🤔

uuf6429: I reran CI for the main branch and it finished successfully (also rerun it for this branch several times and it always failed).

That would only mean that something in this branch is breaking things badly later on.

uuf6429: I did a lot of changes, but somehow I cannot get GH workflow to work reliably.

At this point, it would be great to get #43 merged first to "hide" some of the (duplicate) changes in this PR (after rebasing). That would have the added benefit of decreasing the surface area of the problem (i.e. that mink/session is not involved).

aik099 (changes requested):
Instead of merging changes from the `main` branch do a rebase. This way unrelated newly added changes of the `main` branch won't appear as changed made in this PR (as it is now 😢 ).

tests/Custom/TimeoutTest.php
```diff
@@ -3,79 +3,82 @@
...
-            $session->stop();
-        }
-
-        // Reset the array of timeouts to avoid impacting other tests
-        $driver->setTimeouts([]);
+        $this->driver->setTimeouts([
+            'script' => 30000,
+            'page' => 300000,
+            'implicit' => 0,
+        ]);
```
aik099, line 15: Why you've removed the `resetSessions` method which stopped the session (and reset timeouts) so that on next test in same test case you'll get a fresh driver?

What unstabilities in the test suite you're having, that forced you to rewrite this test case?

aik099 (changes requested):

uuf6429:
> Instead of merging changes from the `main` branch do a rebase. This way unrelated newly added changes of the `main` branch won't appear as changes made in this PR (as it is now 😢 ).
> 
> Or maybe you've made a ton of changes?

I first made the changes here then decided it made sense to have a separate PR. After that PR is merged, I can merge back main or do a rebase, and then the changes will not show up here anymore.

tests/Custom/TimeoutTest.php
uuf6429: I did that when I changed the base testcase, since it's not based on a mink session anymore.

I did it precisely because I believed that `resetSessions` wasn't working/behaving well enough:
1. it does not recreate the driver (e.g. driver state remains as is)
2. it cannot handle drivers created separately (not in a session)
3. I suspected that setting selenium timeout values stays permanent, even after the session is restarted
4. recreating the driver or manually calling `setTimeouts` (with an empty array) effectively doesn't do anything because of the way `applyTimeouts()` has been coded (which is why I put the defaults)

Anyway, a lot of that is mostly speculation. I'd have the other PR merged and then work on a solution here.

aik099: I've approved that other PR. Haven't figured out, that most of code was coming it.

I've also understood the default timeout idea: since you're not creating a new driver instance in each test, then you need to change timeouts on the fly to make it work.

aik099 (changes requested):
I've approved the associated PR. Please merge it and then rebase this one, so that I can see better what needs to be reviewed in this PR.

tests/Custom/TimeoutTest.php
uuf6429: Yes, but sadly it didn't fix the problem.

uuf6429: By the way, since the weekend is almost over.. just wanted to say thanks for all the work and effort you've put in this weekend. @aik099

aik099:
The failed builds on the https://github.com/minkphp/webdriver-classic-driver/actions/runs/11654305762 failed due cURL request to Selenium Server took too long to respond (>3 minutes). That is actually default cURL timeout.

Looks like Selenium Server in a Docker container died.

uuf6429:
Thanks for the tries so far.

The situation is very strange:
- it seems to happen only after setting the page timeout to 500ms multiple times
- it applies to SE3 and SE4
- I _think_ it didn't crash before for Firefox
- Couldn't reproduce it locally at all

aik099:
> Thanks for the tries so far.
> 
> The situation is very strange:
> 
> * it seems to happen only after setting the page timeout to 500ms multiple times
> * it applies to SE3 and SE4
> * I _think_ it didn't crash before for Firefox
> * Couldn't reproduce it locally at all

Very strange indeed.

I've tried combo of TimeoutsTest and WindowTest and they've passed.

Seems like some cumulative effect takes place (e.g. tests in between TimeoutsTest and WindowTest add up to produce super long running time). Maybe we've just added more tests and some critical limit was hit.

Keep in mind, that we don't start/top Selenium Server session multiple times (e.g. Driver::start/Driver::stop), but instead do Driver::reset to close windows and clean cookies.

src/WebdriverClassicDriver.php
stof: PHPUnit 11 has a way to write assertions on deprecations. And the Symfony PHPUnit bridge has a trait that we can use in test cases to write assertions for deprecations using the same API.

tests/Custom/TimeoutTest.php
```diff
@@ -6,18 +6,7 @@
...
-            'script' => 30000,
-            'page' => 300000,
-            'implicit' => 0,
-        ]);
-
-        parent::tearDown();
-    }
-
-    public function testInvalidTimeoutSettingThrowsException(): void
+     public function testInvalidTimeoutSettingThrowsException(): void
```
stof, line 9: this change looks weird to me.

src/WebdriverClassicDriver.php
aik099: Do you mean `Symfony\Bridge\PhpUnit\ExpectDeprecationTrait`, which provides the `expectDeprecation` method?

But due to unknown reasons, an attempt to set timeouts in combination with deprecation checking causes issues for window test https://github.com/minkphp/webdriver-classic-driver/actions/runs/11662801115/job/32469893182 .

tests/Custom/TimeoutTest.php
aik099: Fixed.

aik099:
It seems, that I'm getting closer. But I'm out of ideas.

I've tried removing all of the added code one by one to see if the issues will be fixed (e.g. deprecation checking, setting 500ms timeout, etc.). Nothing helped so far.

The problem always seems to be with `testResizeWindow` (see https://github.com/minkphp/webdriver-classic-driver/actions/runs/11663251677/job/32471294459). It does:

1. visits the page
2. loop for 1sec to wait for the window to be resized
3. checks window dimensions

Page visiting fails:

```
Facebook\WebDriver\Exception\Internal\WebDriverCurlException: Curl error thrown for http POST to /session/de256684ffe02744efbebfa2205ed1b8/url with params: {"url":"
http://host.docker.internal:8002/index.html
"}
   ├                                                                                                                                                                                                                  
   ├ Operation timed out after 180000 milliseconds with 0 bytes received                                                                                                                                              
```

The fact that we saw that issue while developing timeout-changing code seems like a coincidence.

uuf6429:
> The problem always seems to be with `testResizeWindow`

Not always, sometimes it fails after and sometimes before - I noticed that when we had the larger build matrix - but it always some to break somewhere there, which is interesting considering the amount of other successful tests between the timeout test and the window test.

> The fact that we saw that issue while developing timeout-changing code seems like a coincidence.

But nonetheless it has to be related to the timeout tests, right? Everything worked fine without the deprecation test I added in this PR.

----

I noticed the GH action you added...that looks very interesting for debugging. 👍

aik099:
> I noticed the GH action you added...that looks very interesting for debugging.

You've welcome to use it. I think can't proceed any further without real-time debugging with xDebug. Trying to arrange it with https://github.com/nektos/act .

aik099:
I've managed https://github.com/nektos/act to work, but all tests are passing. Seems, that issue can only reproduced in the GitHub Actions.

I've also using built-in PHP 8.3 web server and it's passing locally.

Maybe I'll try to use a built-in PHP web server from a different PHP version. We had some issues with this in the past.

uuf6429:
@aik099 thanks for the investigation. I'll give it a look myself now.

uuf6429:
@aik099 I found the solution and possibly the problem, it seems to be because we do not really run headless in the workflow. After the necessary changes, the test passes without a hitch (ands is also quite fast): https://github.com/minkphp/webdriver-classic-driver/actions/runs/11670274546/job/32494122266

aik099:
@uuf6429 , since in all cases it's a timeout to the local web server (started from bootstrap.php) ocurred, then I've decided to run that web server from GitHub Actions directly. The issue I'm facing is that tests can't connect to `http://host.docker.internal:8002` at all.

I have no idea how to fix this. Maybe I need to map  Please help.

uuf6429:
> The issue I'm facing is that tests can't connect to http://host.docker.internal:8002 at all.

Hmm, I'm not sure... maybe try `MINK_HOST: 0.0.0.0:8002` (instead of localhost)?

But in any case, check my previous message - there shouldn't be any difference in how the web server is started.

aik099:
> @aik099 I found the solution and possibly the problem, it seems to be because we do not really run headless in the workflow. After [the necessary changes](https://github.com/minkphp/webdriver-classic-driver/compare/chore/support-old-page-timeout...bugfix/investigate-crash?expand=1), the test passes without a hitch (~and is also quite fast~ maybe not, it was just a few seconds faster): https://github.com/minkphp/webdriver-classic-driver/actions/runs/11670274546/job/32494122266
> 
> I copied that config from another project, the main noteworthy things is that (1) the docker service does not start xvfb (true headless), (2) the config requests that chrome is started headless and (3) config also requests chrome not to use the gpu.
> 
> On a related subject, as a Behat user, this thing about properly configuring headless has always been a bit of a hassle. Even the slightly wrong config causes crashes and extremely hard to understand errors. Since this driver takes care of some defaults settings, I'm wondering if it may make sense to bake this in to make it easier for the end users. The problem is how do we pass this information (`$headless = true`) from mink?
> 
> One possible way is to pass a non-standard capability (e.g. "headless") which we then handle inside the driver. What do you think?

Here are my configs (in desired capabilities) for headless testing:

```
'moz:firefoxOptions' => array(
    'args' => array(
        '--headless',
        '--width=1550',
        '--height=1160',
    ),
),
'goog:chromeOptions' => array(
    'args' => array(
        '--headless',
        // '--window-size=1550,1160', // doesn't work
        '--start-maximized', // works
    ),
),
```

1. we can look at `'--headless'` element presence in the options sub-array in desired capabilities for a used browser and if found add other args as needed (if they're not provided already)
2. I would keep headless disabled by default to allow developers to see what tests are doing in the web browser
3. I've noticed, that Firefox also isn't executed in the headless mode and maybe it should
4. since the XVFB is disabled now, then maybe we can enable `testWindowMaximize` again (excluded currently due to XVFB usage on GitHub Actions)

tests/Custom/TimeoutTest.php
aik099: > By the way, since the weekend is almost over.. just wanted to say thanks for all the work and effort you've put in this weekend. @aik099

@uuf6429 , my words exactly.

aik099:
According to https://github.com/sebastianbergmann/phpunit/issues/5062 the `exceptDeprecation` method won't work in PHPUnit 10. I'll guess we need to take action once we support PHPUnit 10.

P.S.
It might not be that hard to add PHPUnit 10 support according to the https://github.com/minkphp/driver-testsuite/issues/64 .

aik099:
Next steps:

1. create separate PR with headless testing support
2. maybe support specifying desired capabilities on GitHub Actions side
3. rebase this PR on that PR to see if it actually helped with Google Chrome freezing issue
.

aik099:
> or
> 
> You can skip the 500ms timeout-related test on Google Chrome entirely. For that test, we need to understand if short timeout actually stops long loading pages in every Selenium Server version, BUT no matter in what browser.

@uuf6429 , I've implemented this solution. As you can see it works.

If you agree with proposed solution (to run test added in this PR only in Firefox builds), then please:

1. merge the https://github.com/minkphp/webdriver-classic-driver/pull/47
2. merge this PR

.

uuf6429:
Call me stubborn, but I'd really like an answer for this behaviour. The more I dug, there more I had this creeping sensation that it might be affecting the end users. 🙈 

We could be pragmatic and skip the test as you did and continue the investigation separately.

aik099:
The connected #47 is adjusted according to the review comments. You can proceed with a plan from the https://github.com/minkphp/webdriver-classic-driver/pull/38#issuecomment-2468197711 now.

aik099:
Rebased.

uuf6429:
@aik099 hey, it seems the crash is also happening for firefox/selenium2:
https://github.com/minkphp/webdriver-classic-driver/actions/runs/11803930387/job/32883023966?pr=38
I guess that build was cancelled in your last changes, so we never saw this happening:
https://github.com/minkphp/webdriver-classic-driver/actions/runs/11793213060/job/32848310340

We could also skip this case, but the situation (combination of browser with selenium) is in general very odd.

aik099 (approved):

uuf6429:
Sorry @aik I just decided to exclude the firefox d&d case in this PR, just to get rid of it quick.

aik099:
> Sorry @aik099 I just decided to exclude the firefox d&d case in this PR, just to get rid of it quick and finally have a green check for tests (and since that part can reuse some code from this PR).

Do as you see fit. Don't forget to close the https://github.com/minkphp/webdriver-classic-driver/pull/49 mentioning that fact. 😉

aik099:
@uuf6429 , when merging anything (especially using squash) please adjust the commit message. We have cleaner `git log` with fewer commits now, but these commits have very unclean commit messages, like https://github.com/minkphp/webdriver-classic-driver/commit/ca55cdef2cd5aacfb3c037352c5fa66b51e5a3e6 .

uuf6429:
@aik099 isn't that just the description? Or did you mean the commit message (that it mentions MinkSelenium2Driver)?

https://github.com/minkphp/webdriver-classic-driver/commits/main/

aik099:
Yeah, I meant the commit message itself, which explains how we've got to the code we have, that doesn't matter anymore:

```
Support MinkSelenium2Driver page timeout (https://github.com/minkphp/webdriver-classic-driver/pull/38)
* Support MinkSelenium2Driver page timeout

* Trigger deprecation of old-style page timeout

* Replace duplicate code with comment

* Test deprecations

* Rework custom tests to not depend on magic setup

* Set up standard defaults to avoid contaminated sessions

* [debug] Only run tests that fail

* [debug] Adjusted PHPUnit 10 test filtering expression to work on PHPUnit 9

* [debug] remove irrelevant (for debugging) build config sections to speed it up

* [debug] Reduce matrix size even more

* Don't let driver define Selenium Server default timeouts

* [debug] confirm, that timeout tests executed at the beginning would break window tests down the road

* Revert "[debug] confirm, that timeout tests executed at the beginning would break window tests down the road"

This reverts commit https://github.com/minkphp/webdriver-classic-driver/commit/b33289bebd9138dd5394b3c4c80e334db9f23b28.

* [debug] see if Selenium actually sets timeouts as needed

* Don't reset timeout at the test end, because new driver instance is created in setup

* [debug] PhpStan fix

* [debug] see if session with incorrect timeout is shared between tests

* Fixed formatting error

* [debug] see if deprecation catching causes timeout issues

* [debug] see of sleepy page causes a timeout issue

* [debug] change data provider declaration of the timeouts test

* [debug] removed debug output in the driver code, because not timeouts are causing the issue with window test

* [debug] Look inside GitHub Actions with an SSH session

* [debug] start web server outside of PHPUnit bootstrap file

* [debug] remove unrelated build config parts

* [debug] corrected host of locally started mink server

* [debug] corrected hostname for started web server checking

* [debug] another try to get local web server running

* [debug] reduce test count to figure out why web server is not working

* [debug] Experiment with WEB_FIXTURES_HOST env var detection

* [debug] another try to debug what URL is actually used by test suite

* [debug] another try

* Allow overriding env vars from "phpunit.xml.dist" via env vars from GitHub Actions

* [debug] removed some of the debug code

* [debug] change local web server detection code to avoid it's crashing in the checking process

* [debug] try to make web server available for Selenium Server

* [debug] undo experimental port mapping

* [debug] attempt to start web sever on all ips

* [debug] rolled back CI code that reduced build count

* [debug] rolled back the debug code, that started web server outside of PHPUnit code

* rolled back changes to "phpunit.xml.dist" (to be discussed in other issue)

* Partially skip short "pageLoad" timeout test in Google Chrome

* Fully skip short "pageLoad" timeout test in Google Chrome

* Only skip test added in this PR for Google Chrome

* Apparently "chromium" and "edge" browsers also freeze, when setting "pageLoad" timeout several times. So skip test in there as well.

* Exclude timeout deprecation test only on GitHub Actions

* Minor test fixes and improvements

* Disable test on old firefox

* Fix https://github.com/minkphp/MinkSelenium2Driver/issues/391

---------

Co-authored-by: Alex <aik099@users.noreply.github.com>
```

PR #39, opened 2024-11-02
Wrap WebDriver exception about failed to load page into DriverException

aik099:
Wrap WebDriver's `TimeoutException` and `ScriptTimeoutException` exceptions into `DriverException`, when the page wasn't opened due to `page` timeout being met.

P.S.
Based on https://github.com/minkphp/MinkSelenium2Driver/pull/400 .

### DIFF
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -14,6 +14,8 @@ use Behat\Mink\Driver\CoreDriver;
 use Behat\Mink\Exception\DriverException;
 use Facebook\WebDriver\Exception\NoSuchCookieException;
 use Facebook\WebDriver\Exception\NoSuchElementException;
+use Facebook\WebDriver\Exception\ScriptTimeoutException;
+use Facebook\WebDriver\Exception\TimeoutException;
 use Facebook\WebDriver\Exception\UnsupportedOperationException;
 use Facebook\WebDriver\Exception\WebDriverException;
 use Facebook\WebDriver\Remote\DesiredCapabilities;
@@ -155,7 +157,11 @@ class WebdriverClassicDriver extends CoreDriver
 
     public function visit(string $url): void
     {
-        $this->getWebDriver()->navigate()->to($url);
+        try {
+            $this->getWebDriver()->navigate()->to($url);
+        } catch (TimeoutException|ScriptTimeoutException $e) {
+            throw new DriverException('Page failed to load: ' . $e->getMessage(), 0, $e);
+        }
     }
 
     public function getCurrentUrl(): string
--- a/tests/Custom/TimeoutTest.php
+++ b/tests/Custom/TimeoutTest.php
@@ -35,6 +35,7 @@ class TimeoutTest extends TestCase
         assert($driver instanceof WebdriverClassicDriver);
 
         $this->expectException(DriverException::class);
+        $this->expectExceptionMessage('Invalid timeout type: invalid');
 
         $driver->setTimeouts(['invalid' => 0]);
     }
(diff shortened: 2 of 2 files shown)


src/WebdriverClassicDriver.php
```diff
@@ -155,7 +157,15 @@ public function reset(): void
 
     public function visit(string $url): void
     {
-        $this->getWebDriver()->navigate()->to($url);
+        try {
+            $this->getWebDriver()->navigate()->to($url);
+        } catch (TimeoutException $e) {
+            // The W3C compatible Selenium Server.
+            throw new DriverException('Page failed to load: ' . $e->getMessage(), 0, $e);
+        } catch (ScriptTimeoutException $e) {
```
uuf6429, line 165: 1. I would consider merging those two into one, since the body logic is the same:
   ```php
        } catch (TimeoutExceptio|ScriptTimeoutException $e) {
   ```
2. In other places, we typically catch all exceptions and wrap them in the driver one. I'm wondering if it wouldn't also make sense to do that here too?
   ```php
        } catch (Throwable $e) {
            throw new DriverException('Page failed to load: ' . $e->getMessage(), 0, $e);
        }
   ```

tests/Custom/TimeoutTest.php
```diff
@@ -64,4 +65,17 @@ public function testLongTimeoutWaitsForElementToAppear(): void
...
         $this->assertNotNull($element);
     }
+
+    public function testShortPageLoadTimeoutThrowsException(): void
+    {
+        $session = $this->getSession();
+        $driver = $session->getDriver();
+        \assert($driver instanceof WebdriverClassicDriver);
+
+        $driver->setTimeouts(array('page' => 1)); // Use 1ms timeout to avoid any successful page load.
```
uuf6429, line 75: Please use short array syntax.

tests/Custom/TimeoutTest.php
```diff
@@ -64,4 +65,17 @@ public function testLongTimeoutWaitsForElementToAppear(): void
...
+
+        $this->expectException('\Behat\Mink\Exception\DriverException');
```
uuf6429, line 77: Please use class reference (`DriverException::class`)...string class names are often times forgotten and/or not properly recognized by static code analysers.

tests/Custom/TimeoutTest.php
```diff
@@ -64,4 +65,17 @@ public function testLongTimeoutWaitsForElementToAppear(): void
...
+        $this->expectExceptionMessage('Page failed to load: ');
+        $session->visit('https://www.webpagetest.org/'); // Use external URL, because it loads slower, then local.
```
uuf6429, line 79: I'm not sure I like this test:
1. it is very unlikely, but still remotely possible that the page loads in 1ms.
2. it depends on some arbitrary external site out of our control - all sorts of things might happen there.

May I suggest using our fixtures with a page that intentionally simulates a slow page load?
Something like so:
```php
<?php

// slow-page.php

echo 'Hello ';
sleep(2);
echo 'World';
```

That would mean putting this in the driver-testsuite, which I think would actually make sense and benefit other drivers.

uuf6429 (changes requested):

tests/Custom/TimeoutTest.php
aik099: Done.

src/WebdriverClassicDriver.php
uuf6429: > In what PHP version this syntax was introduced?

[7.1](https://3v4l.org/2smoG)

> I prefer using most specific exception classes to understand when WebDriver misbehaves and starts to use unexpected exception classes.

I would normally agree, but not doing that breaks the interface contract and fixing it later on is technically a BC break. 🙈 

I'm only thinking about it - I'm not fully convinced in either way.

aik099: > ```} catch (TimeoutExceptio|ScriptTimeoutException $e) {```

Done.

> I would normally agree, but not doing it breaks the interface contract and fixing it later on is technically a BC break. 🙈 I'm only thinking about it - I'm not fully convinced in either way.

Let's keep it. WebDriver exception thrown instead of DriverException isn't covered by BC promise anyway 😄

aik099:
You can safely merge this PR if no other discussable topics are left.

src/WebdriverClassicDriver.php
```diff
@@ -155,7 +157,12 @@ public function reset(): void
 
     public function visit(string $url): void
     {
-        $this->getWebDriver()->navigate()->to($url);
+        try {
+            $this->getWebDriver()->navigate()->to($url);
+        } catch (TimeoutException|ScriptTimeoutException $e) {
+            // The W3C compatible Selenium Server.
```
uuf6429, line 163: @aik099 sorry, very small final point: that comment is now a bit irrelevant.

aik099: Done.

PR #41, opened 2024-11-02
Optimize build matrix

### LINKED ISSUE #40: Possibly reduce build matrix size
### Current state

Right now the test matrix looks like this:

1. test `each web browser` (firefox & chrome) x `every PHP version` (7.4, 8.0, 8.1, 8.2) x `every selenium server version` (2, 3, 4) = 24 builds
4. test rarely used web browsers, like chromium and edge = 2 builds

Total: **26 builds**.

### Proposition

If we assume, that the driver doesn't have any browser-specific IFs, then we should:

1. test `each web browser` x `each selenium version` on the latest PHP version = 6 builds
3. test `single browser` x `latest selenium version` on every PHP version = 4 builds
4. test rarely used web browsers, like chromium and edge as now = 2 builds

Total: **12 builds**.

Does this make sense or am I missing something?

uuf6429:
Fixes #40.

GitHub workflows have some limitations on setting up "constants" or "global variables".
There are two solutions:
1. using a placeholder and then replacing it when needed with a sort of ternary operation:
   ```yaml
   ${{ matrix.browser == 'default' && env.DEFAULT_BROWSER || matrix.browser }}
   ```
   It's quite verbose, not so readable and "default" will show up in the CI steps (not so nice).
   See run: https://github.com/minkphp/webdriver-classic-driver/actions/runs/11645388813
2. using a previous step to set up defaults, as [mentioned in the documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-using-an-output-to-define-two-matrices).
   **On the whole this seems to work better** and is less verbose. The extra step might be confusing and initially, it will not properly list all steps until that first one is executed.
   See run: https://github.com/minkphp/webdriver-classic-driver/actions/runs/11645451262

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
           docker compose logs --no-color &> ./logs/selenium.log
 
       - name: Archive logs artifacts
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: ${{ failure() }}
         with:
           name: logs_php-${{ matrix.php }}_selenium-${{ matrix.selenium }}_${{ matrix.browser }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,36 +37,40 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
+    env:
+      DEFAULT_PHP: '8.3'
+      DEFAULT_BROWSER: 'chrome'
+      DEFAULT_SELENIUM: '4'
     strategy:
       matrix:
         # Check different Selenium versions for common browsers
-        php: [ '8.3' ]          # DEFAULT_PHP
+        php: [ 'default' ]
         browser: [ 'firefox', 'chrome' ]
         selenium: [ '2.53.1', '3', '4' ]
         include:
           # Check on different PHP versions
-          - browser: 'chrome'   # DEFAULT_BROWSER
-            selenium: '4'       # DEFAULT_SELENIUM
-            php: '7.4'
-          - browser: 'chrome'   # DEFAULT_BROWSER
-            selenium: '4'       # DEFAULT_SELENIUM
-            php: '8.0'
-          - browser: 'chrome'   # DEFAULT_BROWSER
-            selenium: '4'       # DEFAULT_SELENIUM
-            php: '8.1'
-          - browser: 'chrome'   # DEFAULT_BROWSER
-            selenium: '4'       # DEFAULT_SELENIUM
-            php: '8.2'
-          - browser: 'chrome'   # DEFAULT_BROWSER
-            selenium: '4'       # DEFAULT_SELENIUM
-            php: '8.3'
+          - php: '7.4'
+            browser: 'default'
+            selenium: 'default'
+          - php: '8.0'
+            browser: 'default'
+            selenium: 'default'
+          - php: '8.1'
+            browser: 'default'
+            selenium: 'default'
+          - php: '8.2'
+            browser: 'default'
+            selenium: 'default'
+          - php: '8.3'
+            browser: 'default'
+            selenium: 'default'
           # Check less-common browsers
-          - php: '8.3'          # DEFAULT_PHP
-            selenium: '4'       # DEFAULT_SELENIUM
+          - php: 'default'
             browser: 'edge'
-          - php: '8.3'          # DEFAULT_PHP
-            selenium: '4'       # DEFAULT_SELENIUM
+            selenium: 'default'
+          - php: 'default'
             browser: 'chromium'
+            selenium: 'default'
       fail-fast: false
 
     steps:
@@ -77,7 +81,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           coverage: "xdebug"
-          php-version: "${{ matrix.php }}"
+          php-version: "${{ matrix.php == 'default' && env.DEFAULT_PHP || matrix.php }}"
           ini-file: development
 
       - name: Install dependencies
@@ -86,7 +90,7 @@ jobs:
 
       - name: Start Selenium
         run: |
-          SELENIUM_IMAGE=selenium/standalone-${{ matrix.browser }}:${{ matrix.selenium }} docker compose up --wait
+          SELENIUM_IMAGE=selenium/standalone-${{ matrix.browser == 'default' && env.DEFAULT_BROWSER || matrix.browser }}:${{ matrix.selenium == 'default' && env.DEFAULT_SELENIUM || matrix.selenium }} docker compose up --wait
 
       - name: Wait for selenium to start
         run: |
@@ -94,10 +98,10 @@ jobs:
 
       - name: Run tests
         env:
-          SELENIUM_VERSION: ${{ matrix.selenium }}
+          SELENIUM_VERSION: ${{ matrix.selenium == 'default' && env.DEFAULT_SELENIUM || matrix.selenium }}
           DRIVER_URL: http://localhost:4444/wd/hub
           WEB_FIXTURES_HOST: http://host.docker.internal:8002
-          WEB_FIXTURES_BROWSER: ${{ matrix.browser }}
+          WEB_FIXTURES_BROWSER: ${{ matrix.browser == 'default' && env.DEFAULT_BROWSER || matrix.browser }}
           DRIVER_MACHINE_BASE_PATH: /fixtures/
         run: |
           vendor/bin/phpunit -v --coverage-clover=coverage.xml --colors=always --testdox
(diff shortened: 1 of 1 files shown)


uuf6429:
@aik099 the change in tests has revealed a "bug" in the browser name test where the input is not matching the output (chromium -> chrome and edge -> MicrosoftEdge). Not sure how to solve this TBH. :(

aik099:
You're getting a double build count (on `pull_request` and on `push`) because you're creating branches for your PRs in the `minkphp/webdriver-classic-driver` repo instead of your fork repo `uuf6429/webdriver-classic-driver`.


Here is my setup:

Do once:
1. clone the `minkphp/webdriver-classic-driver` repo (that is `origin` remote)
2. fork the `minkphp/webdriver-classic-driver` repo as `aik099/webdriver-classic-driver`
3. add `aik099` remote that points to the fork

Do each time:
1. create a PR branch
2. make code changes & commits
3. push that branch into `aik099` remote with `--set-upstream` option so they're linked now
4. create a pull request from just pushed branch on the `aik099/webdriver-classic-driver`

You can see for yourself by comparing the build in mine & your PRs:
* mine: https://github.com/minkphp/webdriver-classic-driver/pull/39/checks - only `pull_request` part
* yours: https://github.com/minkphp/webdriver-classic-driver/pull/38/checks - both `push` and `pull_request` part
.

uuf6429:
> You're getting a double build count (on pull_request and on push) because you're creating branches for your PRs in the minkphp/webdriver-classic-driver repo instead of your fork repo uuf6429/webdriver-classic-driver.

I was under the impression that a fork is needed for non-maintainers to contribute. As a maintainer, why should I have a fork? In that case I believe the workflow should be better adjusted.. Isn't it also more convenient for you to work on the main repo directly?

aik099:
> @aik099 the change in tests has revealed a "bug" in the browser name test where the input is not matching the output (chromium -> chrome and edge -> MicrosoftEdge). Not sure how to solve this TBH. :(

1. in the driver constructor just do `$this->browserName = $browserName;`
2. in other places (`\Mink\WebdriverClassicDriver\WebdriverClassicDriver::getBrowserSpecificCapabilities` and `\Mink\WebdriverClassicDriver\WebdriverClassicDriver::initCapabilities` execute the `self::BROWSER_NAME_ALIAS_MAP[$this->browserName] ?? $this->browserName` instead of `$this->browserName`

This way developer will get the same browser as he provided at driver construction time, but the mapping (to satisfy WebDriver) would be purely driver internal stuff.

uuf6429:
Out of curiosity, I checked how [PHPStan does it](https://github.com/phpstan/phpstan/blob/2.0.x/.github/workflows/tests.yml). It looks similar to something I saw in the GH docs, wouldn't this make more sense?
```yaml
on:
  pull_request:
  push:
    branches:
      - "main"
```

.github/workflows/ci.yml
```diff
@@ -10,6 +10,20 @@ defaults:
...
+    outputs:
+      php: ${{ steps.defaults.outputs.php }}
+      browser: ${{ steps.defaults.outputs.browser }}
+      selenium: ${{ steps.defaults.outputs.selenium }}
+    steps:
+      - id: defaults
+        run: |
+          echo 'php="8.3"' >> "$GITHUB_OUTPUT"
+          echo 'browser="chrome"' >> "$GITHUB_OUTPUT"
+          echo 'selenium="4"' >> "$GITHUB_OUTPUT"
```
aik099, line 25: Can this code be simplified according to https://docs.github.com/en/enterprise-server@3.10/actions/writing-workflows/choosing-what-your-workflow-does/setting-a-default-shell-and-working-directory ?

Maybe move them to a specific job instead of above all jobs.

```
jobs:
  job1:
    runs-on: ubuntu-latest
    defaults:
      run:
        shell: bash
        working-directory: ./scripts
```

Then use these defaults as `{{ jobs.<job_id>.defaults.run }}`.

The `$GITHUB_OUTPUT` stuff looks like a workaround to me.

aik099 (changes requested):

aik099:
> Out of curiosity, I checked how [PHPStan does it](https://github.com/phpstan/phpstan/blob/2.0.x/.github/workflows/tests.yml). It looks similar to something I saw in the GH docs, wouldn't this make more sense?
> 
> ```yaml
> on:
>   pull_request:
>   push:
>     branches:
>       - "main"
> ```

Belive me. Having PR branches on the main repo is a bad idea. Even worse these PR branches are forked to the contributors repos as well.

aik099:
> > You're getting a double build count (on pull_request and on push) because you're creating branches for your PRs in the minkphp/webdriver-classic-driver repo instead of your fork repo uuf6429/webdriver-classic-driver.
> 
> I was under the impression that a fork is needed for non-maintainers to contribute. As a maintainer, why should I have a fork? In that case I believe the workflow should be better adjusted.. Isn't it also more convenient for you to work on the main repo directly?

I use the same approach for all repos regardless of the fact, that have `push` access to them (means I'm a maintainer there) 😄 . As I saw @stof does the same in https://github.com/minkphp/webdriver-classic-driver/pull/3 .

aik099:
> Out of curiosity, I checked how [PHPStan does it](https://github.com/phpstan/phpstan/blob/2.0.x/.github/workflows/tests.yml). It looks similar to something I saw in the GH docs, wouldn't this make more sense?
> 
> ```yaml
> on:
>   pull_request:
>   push:
>     branches:
>       - "main"
> ```

I'm not saying no. Let's hear other opinions: @stof , @oleg-andreyev , @mvorisek .

.github/workflows/ci.yml
uuf6429: It has to run at the job level, because those defaults are used in a matrix of a job.

The problem is to use variables in a matrix.. apparently it's not possible.

aik099: As with tests I prefer to do some copy/paste to increase code clarity. If we need to change the default browser/selenium version we can do several `Find & Replace`.

mvorisek:
Thank you for asking me. I am strongly for running CI on all pushes - I very often want to run CI on non-PR branch - like before I submit PR, for experiments, ...

uuf6429:
> Thank you for asking me. I am strongly for running CI on all pushes - I very often want to run CI on non-PR branch - like before I submit PR, for experiments, ...

That sounds like you want to run it manually, right? In that case it would make sense to also keep `on: workflow_dispatch`.

mvorisek:
I simply fork them, enable CI for that repo and push. I prefer CI on every push instead having to trigger it manually.

.github/workflows/ci.yml
uuf6429: The problem with that is that it doesn't work well with short values like '4' or '8.3' (on top of which there are several places that shouldn't be replaced).

Other than that, the code is actually relatively readable (except that first step, but let's be honest, github workflows aren't that readable in general).

aik099:
> I simply fork them, enable CI for that repo and push. I prefer CI on every push instead having to trigger it manually.

@mvorisek , in this scenario you'll also get double builds:

1. when pushing to the forked repo (build in a forked repo on the `push` event)
2. when commit gets added to a pull request (build in origin repo on the `pull_request` event)

The problem, that was supposed to be solved by restricting `push` builds on the `main` branch was double builds, when maintainers create PR branches directly in the repo itself and not on their fork (see https://github.com/minkphp/webdriver-classic-driver/pull/41#issuecomment-2453116465 ).

What do you recommend?

mvorisek:
I am maintainer of atk4 project and we run such CI configuration for years - https://github.com/atk4/ui/blob/5.2.0/.github/workflows/test-unit.yml#L4-L7. So far the CI is great. We even run CI every 2 hours using cron, so we can easily notice dependencies issues.

I found the CI run two times is actually useful, as you can easily notice a) difference of branch vs. rebased branch on PR branch (that's how `pull_request` is tested), b) randomly failing tests.

.github/workflows/ci.yml
```diff
@@ -37,18 +51,47 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
+    needs: defaults
+    env:
+      DEFAULT_PHP: '8.3'
+      DEFAULT_BROWSER: 'chrome'
+      DEFAULT_SELENIUM: '4'
```
aik099, line 58: Is this part needed?

.github/workflows/ci.yml
aik099: I see. I've just found https://partial.solutions/2023/advanced-matrix-testing-with-github-actions.html and the example in there looks very similar to your approach.

Let's keep the current implementation in this PR then.

.github/workflows/ci.yml
aik099, line 58: Is this part needed?

aik099 (changes requested):

.github/workflows/ci.yml
uuf6429: Not at all, good spotting 👍

aik099 (approved):
As for the main PR topic we're all good here. Accepting changes.

As for a side topic (whatever to create PRs from branches in a main repo, that results in double builds) we can create another issue.

I personally suggest creating PRs from repo fork branches.

PR #43, opened 2024-11-03
Simplify custom tests

uuf6429:
The base TestCase from driver-testsuite has some disadvantages:
- it tends to require some additional boilerplate for most driver-specific tests
- it's a bit unclear/convoluted for simple test scenarios
- in principle, tests requiring mink/getSession() are probably misplaced and should exist in the driver testsuite when you think about it

With that in mind, I came up with a simple base TestCase class that copies some of the functionality of the original one.
In my opinion, we should make helper traits out of the original one, which driver authors could use - but that's a bit off topic.

### DIFF
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -956,12 +956,15 @@ class WebdriverClassicDriver extends CoreDriver
                     case 'script':
                         $timeouts->setScriptTimeout($param / 1000);
                         break;
+
                     case 'implicit':
                         $timeouts->implicitlyWait($param / 1000);
                         break;
+
                     case 'page':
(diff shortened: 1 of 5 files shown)


aik099 (approved):
Now I've got the core idea behind this PR: if we want to test the driver, then we need to instantiate driver class and only call it.

Currently, in the driver test suite, we use `NodeElement` and `Session` to manipulate the driver, but that only slows things down and makes it harder to understand what driver calls are actually made. We're basically testing the Mink Driver using the Mink itself, which is a bad practice.

You can suggest a driver test suite rewrite idea at https://github.com/minkphp/driver-testsuite , but I'm not sure who will volunteer to rewrite a test suite, that already works.

PR #47, opened 2024-11-11
Allow skipping custom tests (copy/paste from "driver-testsuite")

aik099:
The problem found during the #38 development.

### DIFF
--- a/tests/Custom/TestCase.php
+++ b/tests/Custom/TestCase.php
@@ -37,4 +37,16 @@ class TestCase extends \PHPUnit\Framework\TestCase
     {
         return WebdriverClassicConfig::getInstance();
     }
+
+    /**
+     * @before
+     *
+     * @return void
+     */
+    protected function checkSkippedTest()
+    {
+        if (null !== $message = self::getConfig()->skipMessage(get_class($this), $this->getName(false))) {
+            $this->markTestSkipped($message);
+        }
+    }
 }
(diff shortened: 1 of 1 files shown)


tests/Custom/TestCase.php
```diff
@@ -37,4 +37,16 @@ protected function getConfig(): WebdriverClassicConfig
     {
         return WebdriverClassicConfig::getInstance();
     }
+
+    /**
+     * @before
+     *
+     * @return void
+     */
+    protected function checkSkippedTest()
```
uuf6429, line 46: Please use `: void` and remove that part from PHPDoc.

tests/Custom/TestCase.php
```diff
@@ -37,4 +37,16 @@ protected function getConfig(): WebdriverClassicConfig
...
+    {
+        if (null !== $message = self::getConfig()->skipMessage(get_class($this), $this->getName(false))) {
```
uuf6429, line 48: I would enclose the assignment in brackets (or put it in a separate line), it would be more readable.
I would also replace `get_class($this)` with `static::class`:
```php
$message = $this->getConfig()->skipMessage(self::class, $this->getName(false));

if ($message !== null) {
    $this->markTestSkipped($message);
}
```

uuf6429 (changes requested):
Just a few minor things.

tests/Custom/TestCase.php
aik099: Done.

tests/Custom/TestCase.php
aik099: Done.

aik099 (commented):
I've corrected both things.

PR #51, opened 2024-11-19
Fix and streamline capability handling

### LINKED ISSUE #46: Condition prevents setting default capabilities
https://github.com/minkphp/webdriver-classic-driver/blob/61b12d05f4185a1f9d0676fc66d559c1e435a97a/src/WebdriverClassicDriver.php#L801
☝️ That condition means that we cannot overwrite capabilities by the php-webdriver - in case of Firefox, because of this:
https://github.com/php-webdriver/php-webdriver/blob/4c18b78d93e4a724ad2dd6427ed658e61a45c569/lib/Remote/DesiredCapabilities.php#L299

For example, the following firefox capability change would not work:
```php
    public const DEFAULT_CAPABILITIES = [
        'default' => [
            'platform' => 'ANY',
            'name' => 'Behat Test',
            'deviceOrientation' => 'landscape',
            'deviceType' => 'desktop',
        ],

        WebDriverBrowserType::FIREFOX => [  // <<<<<<<
            'moz:firefoxOptions' => [
                'args' => [
                    'some-argument',
                ],
            ],
        ],
    ];
```
_(that also applies to the default one; we're not able to overwrite anything predefined in the php-webdriver defaults)_

I think this code was copied from MinkSelenium2Driver, so I never really thought about it.

On a related note, I'm unsure how in general we are supposed to behave - should we be merging config, or generally overwriting it? That applies to php-webdriver defaults, our defaults and the end user's desired capaibilities.

uuf6429:
Fixes #46.

## What Changed

1. Fixes code preventing us from overwriting capbalities when they're already set by php-webdriver.
2. Streamlines the flow so that it is clearer how it behaves:
    1. start with the browser-specific capabilities provided by php-webdriver
    2. add mink default capabilities
    3. add mink browser-specific default capabilities
    4. add capabilities defined by the end user
3. 'add', in this case, always overwrites existing capabilities

## Remaining Questions

- How do we test this? (should we even test it?) In principle this is private behaviour and not something accessible from the outside.
- Should we be worried that this is a BC break? Existing users that have broken config would break their system, when in the past that config was unintentionally ignored.

### DIFF
--- a/tests/Custom/CapabilityTest.php
+++ b/tests/Custom/CapabilityTest.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace Mink\WebdriverClassicDriver\Tests\Custom;
+
+use Mink\WebdriverClassicDriver\WebdriverClassicDriver;
+
+class CapabilityTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * @param array<string, mixed> $desiredCapabilities
+     * @param array<string, mixed> $expectedCapabilities
+     *
+     * @dataProvider capabilitiesDataProvider
+     */
+    public function testThatCapabilitiesAreAsExpected(string $browserName, array $desiredCapabilities, array $expectedCapabilities): void
+    {
+        $driver = $this->createDriverExposingCapabilities($browserName, $desiredCapabilities);
+
+        $this->assertSame($expectedCapabilities, $driver->capabilities);
+    }
+
+    public static function capabilitiesDataProvider(): iterable
+    {
+        yield 'unknown browser starts with default driver capabilities' => [
+            'browserName' => 'fake browser',
+            'desiredCapabilities' => [],
+            'expectedCapabilities' => [
+                'platform' => 'ANY',
+                'name' => 'Behat Test',
+                'deviceOrientation' => 'landscape',
+                'deviceType' => 'desktop',
+            ],
+        ];
+
+        yield 'default capabilities can be customised' => [
+            'browserName' => 'fake browser',
+            'desiredCapabilities' => [
+                'something' => 'custom',
+                'name' => 'Custom Test',
+            ],
+            'expectedCapabilities' => [
+                'platform' => 'ANY',
+                'name' => 'Custom Test',
+                'deviceOrientation' => 'landscape',
+                'deviceType' => 'desktop',
+                'something' => 'custom',
+            ],
+        ];
+
+        yield 'browser-specific default capabilities are added' => [
+            'browserName' => 'chrome',
+            'desiredCapabilities' => [],
+            'expectedCapabilities' => [
+                'browserName' => 'chrome',
+                'platform' => 'ANY',
+                'name' => 'Behat Test',
+                'deviceOrientation' => 'landscape',
+                'deviceType' => 'desktop',
+                'goog:chromeOptions' => [
+                    'excludeSwitches' => ['enable-automation'],
+                ],
+            ],
+        ];
+
+        yield 'browser-specific default capabilities can be customised' => [
+            'browserName' => 'chrome',
+            'desiredCapabilities' => [
+                'name' => 'Custom Test',
+                'goog:chromeOptions' => ['args' => ['a', 'b', 'c']],
+            ],
+            'expectedCapabilities' => [
+                'browserName' => 'chrome',
+                'platform' => 'ANY',
+                'name' => 'Custom Test',
+                'deviceOrientation' => 'landscape',
+                'deviceType' => 'desktop',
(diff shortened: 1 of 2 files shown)


aik099 (changes requested):
> How do we test this? (should we even test it?) In principle this is private behaviour and not something accessible from the outside.

Yes. We should test, that the user can override this (no matter if it came from the `php-webdriver/php-webdriver` library or Mink itself):

* any top-level desired capability (e.g. `platform`)
* any sub-level desired capability (e.g. `goog:chromeOptions > binary`)

> Should we be worried that this is a BC break? Existing users that have broken config would break their system, when in the past that config was unintentionally ignored.

It's not a BC break.

We had similar discussions in the past about the `$name` argument of the `switchIFrame` method. According to the method's DocBlock, it should be always a string. However, we're not checking that and passing that `$name` argument directly to a WebDriver call. This allowed to have bonus undocumented behavior of switching frames by index.

When string argument typing was introduced people started complaining that this is a BC break, but it isn't because we've never advertised that passing frame index is even a supported behavior.

Same here: driver is supported to properly merge desired capabilities and the user is supposed to provide valid desired capabilities. If either of them doesn't do their work, then it's a bug and not a BC break 😄

aik099:
Since we're not mocking the underlying WebDriver library to confirm, that proper desired capabilities were given at session creation time, then we should fall back to a functional test (like with timeouts) to indirectly confirm, that the correct desired capabilities were given.

@uuf6429 , any ideas?

stof:
thus, this driver does not have a stable release yet. So there is no BC guarantee yet.

uuf6429:
> Since we're not mocking the underlying WebDriver library to confirm, that proper desired capabilities were given at session creation time

Why not? I can simply use an anonymous subclass to test that.

uuf6429:
side-note: Currently there's no way to unset top-level capabilities, except with potentially wrong values (e.g. to "unset" `'name'` one would need to pass `['name' => null]` - which is more like invalidation than unsetting, and might still affect selenium).

uuf6429:
@aik099 small reminder about this PR. I believe it's not missing anything, right?

tests/Custom/CapabilityTest.php
```diff
@@ -0,0 +1,103 @@
...
+             * @var null|array<string, mixed>
+             */
+            public ?array $capabilities = null;
+
+            public function __construct(string $browserName, array $desiredCapabilities)
+            {
+                parent::__construct($browserName, $desiredCapabilities);
+
+                $this->capabilities = $this->getDesiredCapabilities();
+            }
```
aik099, line 100: Or you can just make the `getDesiredCapabilities` method public in this extended class.

tests/Custom/CapabilityTest.php
```diff
@@ -0,0 +1,103 @@
...
+
+namespace Mink\WebdriverClassicDriver\Tests\Custom;
+
+use Mink\WebdriverClassicDriver\WebdriverClassicDriver;
+
+class CapabilityTest extends \PHPUnit\Framework\TestCase
+{
+    /**
+     * @param array<string, mixed> $desiredCapabilities
+     * @param array<string, mixed> $expectedCapabilities
```
aik099, line 11: Is this correct behavior to specify only some of the method parameters in the PhpDoc?

aik099 (changes requested):
Looks good otherwise.

tests/Custom/CapabilityTest.php
uuf6429: Yes - PHPDoc is to augment or narrow existing information. E.g. in this case adding a line for `string $browserName` is quite redundant (unless we want to add something special to that parameter - such as a description or `non-empty-string` instead of `string`..).

PHP-CS-Fixer even has a rule/fixer for this: [no_superfluous_phpdoc_tags](https://cs.symfony.com/doc/rules/phpdoc/no_superfluous_phpdoc_tags.html).

tests/Custom/CapabilityTest.php
uuf6429: True, that should also work.. 🤔

uuf6429: Ah, I know why I did it like that - it's because in PHPStan, I cannot type hint a method on an anonymous object (see line 85).

aik099: OK.

tests/Custom/CapabilityTest.php
aik099: OK.

aik099 (approved):

aik099:
To fix the tests you need to rebase.

PR #52, opened 2024-12-02
Normalize behavior of "getText" method

aik099:
See https://github.com/minkphp/driver-testsuite/issues/98.

### DIFF
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -296,11 +296,11 @@ class WebdriverClassicDriver extends CoreDriver
         #[Language('XPath')]
         string $xpath
     ): string {
-        return str_replace(
-            ["\r\n", "\r", "\n"],
+        return trim(str_replace(
+            ["\r\n", "\r", "\n", ' '],
             ' ',
             $this->getElementDomProperty($this->findElement($xpath), 'innerText')
-        );
+        ));
     }
 
     public function getHtml(
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -297,7 +297,7 @@ class WebdriverClassicDriver extends CoreDriver
         string $xpath
     ): string {
         return trim(str_replace(
-            ["\r\n", "\r", "\n", ' '],
+            ["\r\n", "\r", "\n", "\xc2\xa0"],
             ' ',
             $this->getElementDomProperty($this->findElement($xpath), 'innerText')
         ));


src/WebdriverClassicDriver.php
```diff
@@ -296,11 +296,11 @@ public function getText(
         #[Language('XPath')]
         string $xpath
     ): string {
-        return str_replace(
-            ["\r\n", "\r", "\n"],
+        return trim(str_replace(
+            ["\r\n", "\r", "\n", ' '],
```
uuf6429, line 300: Is the `' '` a non-breaking space? I think I would rather use an escaped sequence (`"\xA0"`), otherwise it might not be obvious.

oleg-andreyev: @uuf6429 was writting the same, but if it's NBSP (non-breaking space) it should be U+202F

so it could
```php
mb_chr(8239)
``` 

or 

```php
chr(0xE2) . chr(0x80) . chr(0xAF)
```

no?

aik099: Done. However, I had to use `"\xc2\xa0"`, because both `"\xA0"` and `chr(0xA0)` produced a broken UTF-8 symbol: `�`.

The https://stackoverflow.com/questions/40724543/how-to-replace-decoded-non-breakable-space-nbsp helped.

aik099: Both editors I've had (PhpStorm and Sublime Text) were showing `0xa0` and not `0xc20xa0`:
PhpStorm_NBSP
SublimeText_NBSP

In PHP you can't get it working without adding `\xc2` first. I wonder if it works in Selenium code at all without `\xC2`.

uuf6429: I think I'm fine with it as long as the tests pass...should we maybe wait until the driver-testsuite one is merged and then rerun this?

uuf6429 (approved):

src/WebdriverClassicDriver.php
aik099: > I think I'm fine with it as long as the tests pass...should we maybe wait until the driver-testsuite one is merged and then rerun this?

Done.

See https://github.com/minkphp/webdriver-classic-driver/actions/runs/12132940669/job/33827642410?pr=52:

GitHubActions_GetTextTest

aik099:
@uuf6429 , please do a bugfix release in the foreseeable future (this month) if possible.

PR #53, opened 2024-12-12
Disable Chrome code signing, when used on Mac

aik099:
When executing tests in the Google Chrome (or any Chromium-based Web Browser) on Mac you'll notice you're temp folder starts to grow faster than a lightning strikes.

This happens, because upon launching a Web Browser copies itself (whole 300MB folder) into a temp folder. When the Web Browser exists that temp folder isn't deleted.

Since numerous Selenium Server sessions are started/stopped during test suite run you'll get numerous of ~300MB folders residing under a temp folder.

Solution:

* either use ChromeDriver 131.0.6778.70+ (the issue is fixed in https://issues.chromium.org/issues/379125944);
* or specify `--disable-features=MacAppCodeSignClone` as Chrome arguments in WebDriver configuration.

This PR does the later.

P.S.
@uuf6429, I can't seem to get it working. Maybe related to the bug solved in the https://github.com/minkphp/webdriver-classic-driver/pull/51 .

### DIFF
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -47,6 +47,10 @@ class WebdriverClassicDriver extends CoreDriver
             'goog:chromeOptions' => [
                 // disable "Chrome is being controlled.." notification bar
                 'excludeSwitches' => ['enable-automation'],
+                'args' => array(
+                    // disable browser folder cloning on Mac, when tests are executed
+                    'disable-features=MacAppCodeSignClone',
+                ),
             ],
         ],
 


aik099:
Tested Chromedriver 131.0.6778.139 and the issue is still there. Even though https://issues.chromium.org/issues/379125944 reported (if I'm reading that page correctly), that issue was fixed in the 131.0.6778.70. 😞

uuf6429:
I'm using other arguments in the same way you did it (e.g. `args: ["headless=new"]` for `--headless=new`), so I can confidently say that part works. 🤔

Could you maybe run it and then either inspect selenium settings in `http://localhost:4444/ui` or the chrome settings with `chrome://version/`?

oleg-andreyev:
This is some sort of security/sandboxing , are we sure we want to disable it? I'd rather document it. MacOS itself should handle cleanup, plus how many people or running Chromedriver on MacOS for CI.

aik099:
> Could you maybe run it and then either inspect selenium settings in http://localhost:4444/ui or the chrome settings with chrome://version/? (you should be able to see the CLI args)

@uuf6429 , I've found out why driver defaults weren't used. I'm passing this to allow chromedriver to operate on non-Chrome (Chromium-based) Web Browser:

```php
'goog:chromeOptions' => array(
    'binary' => '/Applications/Brave Browser.app/Contents/MacOS/Brave Browser'
),
```

and that completely replaced driver-provided defaults:

```php
'goog:chromeOptions' => [
        // disable "Chrome is being controlled.." notification bar
        'excludeSwitches' => ['enable-automation'],
        'args' => array(
            // disable browser folder cloning on Mac, when tests are executed
            'disable-features=MacAppCodeSignClone',
        ),
 ],
```
. The mix-and-match approach and on/off switch, as @oleg-andreyev proposed, for default desired capabilities is an interesting thing to discuss.

> In general, I feel like we should have some sort of mechanism (or a sub-project / sub-driver?) to handle capabilities better. For example, it would be nice for our users to just pass a "enable_headless" config which we translate to the right set of capabilities for the specified browser.

@oleg-andreyev , we're discussing this a bit in the #51 . Probably we can have `mink:browserOptions` sub-array in desired capabilities, that is empty by default and user can enable/disable flags, that conceptually enable/disable feature sub-sets of desired capabilities, like `headless=>true`, `honor_browser_defaults=>true`, `honor_driver_defaults=>true`, etc.

> This is some sort of security/sandboxing , are we sure we want to disable it? I'd rather document it. MacOS itself should handle cleanup, plus how many people or running Chromedriver on MacOS for CI.

@oleg-andreyev , I have no idea what this functionality is (probably Chromium source code at https://source.chromium.org/search?q=MacAppCodeSignClone&sq=&ss=chromium can reveal it), but it surely eats a lot of disk space pretty quickly and since on macOS you don't reboot quite often to auto-clear temp folder (as you do on Windows) you can run out of disk space soon enough.

Before creating a PR I prefer to execute the whole Selenium-based test suite locally to see if it passes and only then let GitHub Actions run it. This would save me time waiting for GitHub Actions to crash after some time when I can see the same crash locally and already start to fix things.

aik099:
> This is some sort of security/sandboxing , are we sure we want to disable it? I'd rather document it. MacOS itself should handle cleanup, plus how many people or running Chromedriver on MacOS for CI.

The comments in https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/mac/code_sign_clone_manager.h?q=MacAppCodeSignClone&ss=chromium explain a lot why this was done. If I've understood correctly, then it was done for updating to new version purposes and there is a background process to delete the cloned folder. My guess is that this background process doesn't work or is stopped as long as Selenium-started browser exists as part of `driver.quit()` Selenium command.

I have no idea really why the clones are no deleted, when they should.

aik099:
I've created https://github.com/minkphp/webdriver-classic-driver/issues/54 for discussing general problems with desired capability handling.

PR #55, opened 2025-01-01
Upgrade phpstan; increase phpstan level; fix a few bugs

uuf6429:
- [x] ⚠️ Must be rebased on main after https://github.com/minkphp/webdriver-classic-driver/pull/58 is merged.

### DIFF
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -3,7 +3,7 @@ includes:
 	- vendor/phpstan/phpstan-phpunit/rules.neon
 
 parameters:
-	level: 8
+	level: 9
 	paths:
 		- src
 		- tests
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -3,7 +3,7 @@ includes:
 	- vendor/phpstan/phpstan-phpunit/rules.neon
 
 parameters:
-	level: 9
+	level: 10
 	paths:
 		- src
 		- tests
@@ -25,4 +25,14 @@ parameters:
 			message: '#^Parameter \#1 \$seconds of method Facebook\\WebDriver\\WebDriverTimeouts\:\:setScriptTimeout\(\) expects int, float\|int given\.$#'
 			identifier: argument.type
 			count: 1
+			path: src/WebdriverClassicDriver.php
+		-
+			message: '#^Method Mink\\WebdriverClassicDriver\\WebdriverClassicDriver\:\:getWindowHandleFromName\(\) should return string but returns mixed\.$#'
+			identifier: return.type
+			count: 1
+			path: src/WebdriverClassicDriver.php
+		-
+			message: '#^Parameter \#1 \$handle of method Facebook\\WebDriver\\Remote\\RemoteTargetLocator\:\:window\(\) expects string, mixed given\.$#'
+			identifier: argument.type
+			count: 3
 			path: src/WebdriverClassicDriver.php
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -1,11 +1,28 @@
-parameters:
-    level: 8
-    paths:
-        - src
-        - tests
-    checkMissingIterableValueType: false
-    treatPhpDocTypesAsCertain: false
-
 includes:
-    - vendor/phpstan/phpstan-phpunit/extension.neon
-    - vendor/phpstan/phpstan-phpunit/rules.neon
+	- vendor/phpstan/phpstan-phpunit/extension.neon
+	- vendor/phpstan/phpstan-phpunit/rules.neon
+
+parameters:
+	level: 8
+	paths:
+		- src
+		- tests
+	ignoreErrors:
+		-
+			# See: https://github.com/php-webdriver/php-webdriver/pull/1120
+			message: '#^Parameter \#1 \$seconds of method Facebook\\WebDriver\\WebDriverTimeouts\:\:implicitlyWait\(\) expects int, float\|int given\.$#'
+			identifier: argument.type
+			count: 1
+			path: src/WebdriverClassicDriver.php
+		-
+			# See: https://github.com/php-webdriver/php-webdriver/pull/1120
+			message: '#^Parameter \#1 \$seconds of method Facebook\\WebDriver\\WebDriverTimeouts\:\:pageLoadTimeout\(\) expects int, float\|int given\.$#'
+			identifier: argument.type
+			count: 1
+			path: src/WebdriverClassicDriver.php
+		-
+			# See: https://github.com/php-webdriver/php-webdriver/pull/1120
+			message: '#^Parameter \#1 \$seconds of method Facebook\\WebDriver\\WebDriverTimeouts\:\:setScriptTimeout\(\) expects int, float\|int given\.$#'
+			identifier: argument.type
+			count: 1
+			path: src/WebdriverClassicDriver.php
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -788,11 +788,6 @@ class WebdriverClassicDriver extends CoreDriver
         throw new DriverException('Base driver has not been created');
     }
 
-    protected function getDesiredCapabilities(): array
-    {
-        return $this->desiredCapabilities->toArray();
-    }
-
     // </editor-fold>
 
     // <editor-fold desc="Private Utilities">
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -670,10 +670,6 @@ class WebdriverClassicDriver extends CoreDriver
         $this->getWebDriver()->executeScript($script);
     }
 
-    /**
-     * {@inheritdoc}
-     * @return mixed
-     */
     public function evaluateScript(
         #[Language('JavaScript')]
         string $script
@@ -971,7 +967,7 @@ class WebdriverClassicDriver extends CoreDriver
      * @throws DriverException
      */
     private function executeJsOnElement(
-        RemoteWebElement $element,
+        WebDriverElement $element,
         #[Language('JavaScript')]
         string $script
     ) {
@@ -1044,7 +1040,7 @@ class WebdriverClassicDriver extends CoreDriver
         }
     }
 
-    private function clickOnElement(RemoteWebElement $element): void
+    private function clickOnElement(WebDriverElement $element): void
     {
         $element->getLocationOnScreenOnceScrolledIntoView();
         $element->click();
@@ -1108,14 +1104,11 @@ class WebdriverClassicDriver extends CoreDriver
      */
     private function findElement(
         #[Language('XPath')]
-        string $xpath,
-        ?RemoteWebElement $parent = null
+        string $xpath
     ): RemoteWebElement {
         try {
             $finder = WebDriverBy::xpath($xpath);
-            return $parent
-                ? $parent->findElement($finder)
-                : $this->getWebDriver()->findElement($finder);
+            return $this->getWebDriver()->findElement($finder);
         } catch (\Throwable $e) {
             throw new DriverException("Failed to find element: {$e->getMessage()}", 0, $e);
         }
@@ -1124,7 +1117,7 @@ class WebdriverClassicDriver extends CoreDriver
     /**
      * @throws DriverException
      */
-    private function selectRadioValue(RemoteWebElement $element, string $value): void
+    private function selectRadioValue(WebDriverElement $element, string $value): void
     {
         try {
             (new WebDriverRadios($element))->selectByValue($value);
@@ -1142,7 +1135,7 @@ class WebdriverClassicDriver extends CoreDriver
     /**
      * @throws DriverException
      */
-    private function selectOptionOnElement(RemoteWebElement $element, string $value, bool $multiple = false): void
+    private function selectOptionOnElement(WebDriverElement $element, string $value, bool $multiple = false): void
     {
         try {
             $select = new WebDriverSelect($element);
@@ -1172,7 +1165,7 @@ class WebdriverClassicDriver extends CoreDriver
      *
      * @throws DriverException
      */
-    private function deselectAllOptions(RemoteWebElement $element): void
+    private function deselectAllOptions(WebDriverElement $element): void
     {
         try {
             (new WebDriverSelect($element))->deselectAll();
@@ -1190,7 +1183,7 @@ class WebdriverClassicDriver extends CoreDriver
      * @throws DriverException
      */
     private function ensureInputType(
-        RemoteWebElement $element,
+        WebDriverElement $element,
         #[Language('XPath')]
         string $xpath,
         string $type,
@@ -1233,7 +1226,7 @@ class WebdriverClassicDriver extends CoreDriver
      * @param mixed $value
      * @throws DriverException
      */
-    private function setElementDomProperty(RemoteWebElement $element, string $property, $value): void
+    private function setElementDomProperty(WebDriverElement $element, string $property, $value): void
     {
         $this->executeJsOnElement(
             $element,
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -280,7 +280,7 @@ class WebdriverClassicDriver extends CoreDriver
 
     public function getWindowName(): string
     {
-        $name = (string)$this->evaluateScript('window.name');
+        $name = $this->getAsString($this->evaluateScript('window.name'), 'Window name');
 
         if ($name === '') {
             $name = self::W3C_WINDOW_HANDLE_PREFIX . $this->getWebDriver()->getWindowHandle();
@@ -317,7 +317,7 @@ class WebdriverClassicDriver extends CoreDriver
         return trim(str_replace(
             ["\r\n", "\r", "\n", "\xc2\xa0"],
             ' ',
-            $this->getElementDomProperty($this->findElement($xpath), 'innerText')
+            $this->getAsString($this->getElementDomProperty($this->findElement($xpath), 'innerText'), 'The element\'s innerText')
         ));
     }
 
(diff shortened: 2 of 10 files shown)


phpstan.dist.neon
```diff
@@ -1,10 +1,12 @@
 parameters:
-    level: 8
+    level: 10
     paths:
         - src
         - tests
-    checkMissingIterableValueType: false
```
uuf6429, line 6: note: This option has been removed, plus we probably don't really want it anyway.

phpstan.dist.neon
```diff
@@ -1,10 +1,12 @@
 parameters:
-    level: 8
+    level: 10
```
uuf6429, line 2: 💥

src/WebdriverClassicDriver.php
```diff
@@ -1227,5 +1240,21 @@ private function getElementDomProperty(RemoteWebElement $element, string $proper
         }
     }
 
+    /**
+     * @param mixed $value
+     * @throws DriverException
+     */
+    private function getAsString($value, string $name): string
```
aik099, line 1247: Why this method is needed? Previously we trusted JavaScript to return the correct value, but now we don't.

aik099 (changes requested):

src/WebdriverClassicDriver.php
uuf6429: PHPStan (correctly) warns that we may get values that are not stringable (there's a test for that too), in which case all sorts of weird things can happen (in some cases a type error is thrown, in other cases we break the contract with the library user and provide the wrong type).

In any case, there are 4-5 cases where that method is useful, so it makes sense to avoid repeating the same kind of check.

aik099: I understand now. Thank you.

aik099 (approved):

src/WebdriverClassicDriver.php
```diff
@@ -924,12 +925,15 @@ private function charToSynOptions($char, ?string $modifier = null): string
...
+     * ```
+     * $this->executeJsOnXpath($xpath, 'return argument[0].childNodes.length');
+     * ```
+     *
      * @param string $xpath the xpath to search with
      * @param string $script the script to execute
-     *
      * @return mixed
      * @throws DriverException
-     * @example $this->executeJsOnXpath($xpath, 'return argument[0].childNodes.length');
```
uuf6429, line 919: Note that according to PHPDoc standard, `@example` tags are supposed to point to source code files, and not be the example itself (which makes the tag utterly useless in most cases, but anyway).

src/WebdriverClassicDriver.php
```diff
@@ -654,10 +659,6 @@ public function executeScript(
         $this->getWebDriver()->executeScript($script);
     }
 
-    /**
-     * {@inheritdoc}
-     * @return mixed
```
uuf6429, line 648: Parent already has `@return mixed`, so this was redundant.

tests/Custom/CapabilityTest.php
```diff
@@ -80,18 +86,20 @@ public static function capabilitiesDataProvider(): iterable
...
+    private function createDriverExposingCapabilities(string $browserName, array $desiredCapabilities): WebdriverClassicDriver
     {
         return new class($browserName, $desiredCapabilities) extends WebdriverClassicDriver {
             /**
              * @var array<string, mixed>
              */
             public array $capabilities;
 
+            /**
+             * @param array<string, mixed> $desiredCapabilities
```
uuf6429, line 101: Note that we cannot use TCapabilities here, or on line 96 at the moment, because of https://github.com/phpstan/phpstan/issues/9346.

aik099: Maybe you can create a regular (not anonymous) class and just auto-load it in the test.

uuf6429: At this point it's causing more trouble than being convenient, good point.

stubs/RemoteWebDriver.stub
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Facebook\WebDriver\Remote;
+
+interface RemoteWebDriver
```
aik099, line 5: Why these stubs are necessary? Isn't PhpStan clever enough to locate the mentioned class using the Composer's autoloader and use it?

aik099 (changes requested):

stubs/RemoteWebDriver.stub
uuf6429: In all cases it's because the original classes are not defining specific types (or they define the wrong types in one case).

For example, `getWindowHandles` in the original class has [this documentation](https://github.com/php-webdriver/php-webdriver/blob/998e499b786805568deaf8cbf06f4044f05d91bf/lib/Remote/RemoteWebDriver.php#L314):
```php
 * @return array An array of string containing all available window handles.
```

PHPStan only sees `array`, so doing something like `substr($driver->getWindowHandles()[0])` will cause a warning that the value is `mixed`, not a `string`.

The stub is used to overwrite stuff in the original class, which is also why it is an incomplete implementation (and only defines what needs to be changed).

Ideally we should get rid of these stubs at some point, because if the original classes change drastically (e.g. `getWindowHandles` is removed, or made to return `int`'s), PHPStan will not show compatibility warnings on our end.

uuf6429: We have these options:
1. wait until all the stub changes are applied to php-webdriver
   - con: it will probably take some time until that happens
   - pro: least changes on our end
2. merge in the stubs and remove them later when php-webdriver is updated
   - con: double work on our end (to remove the stubs later)
   - con: BC breaking changes in the php-webdriver will not cause warnings
   - pro: very fast release
3. instead of stubs, add the warnings to the [phpstan baseline](https://phpstan.org/user-guide/baseline)
   - con: also double work
   - pro: any changes in the situation (e.g. the warning changes or is gone) will cause a different warning (so that we will be warned in case of BC breaks or when the problem is solved)

aik099: If I understood your point, then the problem cause is a 3rd party library (the `php-webdriver` in this case). It can have method DocBlocks with types that do not match actually returned types. When these types are used in our driver, then PhpStan starts to complain.

My thoughts:

* Stubs (used in point 2) mask the 3rd party library code to look correctly without actually guaranteeing that its behavior matches our stubs.
* By using phpstan baseline warnings (point 3) we expose the problem of 3rd party library and don't need to create a workaround (like in point 2) to make phpstan happier.

I'm not a phpstan expert, but architectural-wise point 3 looks better to me.

uuf6429: I'll go for point 3... will do that later on today (when I'll also submit changes to php-webdriver), thanks.

tests/Custom/WebDriverTest.php
```diff
@@ -72,4 +74,31 @@ public function testClassicDriverCanProvideBrowserName(): void
             $this->driver->getBrowserName()
         );
     }
+
+    public function testThatDriverCatchesUnexpectedAttributeValueType(): void
+    {
+        $driver = $this->createPartialMock(WebdriverClassicDriver::class, ['createWebDriver', 'isStarted', 'getWebDriver', 'executeJsOnXpath']);
```
aik099, line 80: Architecture-wise partial mocks clearly signal that a single class has multiple responsibilities and can be split into several classes.

Here is my suggestion (considering that you can't create a WebDriver without starting a session immediately):

1. in the driver constructor:
    * accept an optional callback, that will have `$seleniumServerUrl` and `array $desiredCapabilities` arguments
    * if no callback is provided, then construct a callback on the fly with `return RemoteWebDriver::create($seleniumServerUrl, $desiredCapabilities);` code
    * store that callback into a new `$createWebDriverCallback` property
2. in the `WebdriverClassicDriver::createWebDriver` method instead of calling `RemoteWebDriver::create` method call the `$this->createWebDriverCallback` callback with the same arguments
3. in this test (and probably other tests using partial mocks) you can create a driver by providing a callback, that will return a WebDriver mock

aik099 (changes requested):

phpstan.dist.neon
```diff
@@ -1,14 +1,41 @@
...
+		-
+			message: '#^Method Mink\\WebdriverClassicDriver\\WebdriverClassicDriver\:\:getDesiredCapabilities\(\) should return array\<string, mixed\> but returns array\.$#'
+			identifier: return.type
+			count: 1
+			path: src/WebdriverClassicDriver.php
+		-
+			message: '#^Method Mink\\WebdriverClassicDriver\\WebdriverClassicDriver\:\:getWindowHandleFromName\(\) should return string but returns mixed\.$#'
+			identifier: return.type
+			count: 1
+			path: src/WebdriverClassicDriver.php
```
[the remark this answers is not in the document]
uuf6429: Because:
- we defined that `getWindowHandleFromName` must return a string (which is correct)
- BUT `getWindowHandles()` in the php-webdriver has an `array` return type (in PHPDoc)...
- ...even though the PHPDoc return description says that it is "an array of strings"
- so PHPStan assumes that it returns an array of mixed
- and in our code we are not checking that the value is indeed a string. *️⃣

In this case, it would be very unlikely that we get something other than an array of strings (since the PHPDoc says so), so I don't think it makes sense to call getAsString() on the result.

Here's the code with some comments:
```php
    private function getWindowHandleFromName(string $name): string // <- must be string
    {
// snip
            foreach ($this->getWebDriver()->getWindowHandles() as $handle) { // $handle is `mixed`, because of bad PHPDoc type in the library
                $this->getWebDriver()->switchTo()->window($handle); // we could have a warning here, if `window()` only accepts strings
                if ($this->evaluateScript('window.name') === $name) { // *️⃣
                    return $handle; // <- we will definitely get a warning here
                }
            }
// snip
    }
```
*️⃣ One simple fix is to add ` && is_string($handle)`, but that is a bit pointless and becomes redundant when the library defines the return type properly.

tests/bootstrap.php
```diff
@@ -4,9 +4,11 @@
...
 use Symfony\Component\Process\Process;
 
-$minkTestServerPort = isset($_SERVER['WEB_FIXTURES_HOST'])
-    ? parse_url($_SERVER['WEB_FIXTURES_HOST'], PHP_URL_PORT)
-    : '8002';
+$fixturesHost = $_SERVER['WEB_FIXTURES_HOST'] ?? '//host:8002';
+if (!is_string($fixturesHost)) {
+    throw new RuntimeException('The fixtures host must be specified in $_SERVER[WEB_FIXTURES_HOST] as a string');
+}
+$minkTestServerPort = parse_url($fixturesHost, PHP_URL_PORT);
```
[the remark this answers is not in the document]
uuf6429: I don't get how that makes it simpler.. it's just skipping parsing of //host:8002, which to me is not only clearer, but also gives a very fast impression of what the value should look like.

That said, I just realised that if the value is not a valid uri, both of our codes would end up with `$minkTestServerPort = false`.

So maybe the easiest way would be:
```php
$fixturesHost = $_SERVER['WEB_FIXTURES_HOST'] ?? '';
$minkTestServerPort = parse_url(is_string($fixturesHost) ? $fixturesHost : '', PHP_URL_PORT) ?: '8002';
```

tests/Custom/WebDriverTest.php
uuf6429: I thought about that in the past, but was worried it would be considered "overengineered", but I do see advantages to this approach, and not just from testing perspective.

phpstan.dist.neon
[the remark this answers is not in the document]
uuf6429: The `getDesiredCapabilities` warning happened because we call `toArray()` of the library's `DesiredCapabilities` class, which again is defined as returning a simple `array` (which PHPStan sees as maybe containing `int` keys, even though we know it is not possible to have integer-keyed capabilities).

----

Anyway, I removed this part entirely because it didn't make sense to convert it to an array - php-webdriver converts the array back into a `DesiredCapabilities` object anyway.

uuf6429:
Sidenote: Since we are now exposing a way to create the internal/library driver, I thought it would make sense to depend on an interface instead of a concretion, however it was not possible because:
1. There are multiple interface involved ([WebDriver](https://github.com/php-webdriver/php-webdriver/blob/main/lib/WebDriver.php) and [JavaScriptExecutor](https://github.com/php-webdriver/php-webdriver/blob/main/lib/JavaScriptExecutor.php))
  - while we cannot use [PHP intersection types](https://php.watch/versions/8.1/intersection-types) yet, we could have nevertheless used it in PHPDoc (and it would be relatively safe thanks to PHPStan)
2. Some functionality (for example `action()` and `isW3cCompliant()`) is not covered by an interface
  - we would need to do expensive runtime checks to ensure that the instantiated driver has those methods (or `assert()` checks to pass PHPStan, but risk calling undefined methods when assertions are disabled).
3. The situation is similar with `RemoteWebElement` (concretion) and `WebDriverElement` (interface) - we use `getDomProperty()` which is not in any interface.

I'll try to suggest some changes to the library.

tests/bootstrap.php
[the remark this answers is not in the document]
aik099: I like the later code, but please replace `''` with `'https://localhost'` (or any valid URL with protocol and hostname), because `parse_url` documentation doesn't guarantee, that the behavior of parsing invalid URLs would be consistent over time.

tests/Custom/WebDriverTest.php
aik099: This PR now has code, that is out of its scope:

1. code about removing anonymous class
2. this code we're talking about here

I'm suggesting moving them into separate PRs with corresponding tests attached.

aik099:
> Sidenote: Since we are now exposing a way to create the internal/library driver, I thought it would make sense to depend on an interface instead of a concretion, however it was not possible because:
> 
> 1. There are multiple interface involved ([WebDriver](https://github.com/php-webdriver/php-webdriver/blob/main/lib/WebDriver.php) and [JavaScriptExecutor](https://github.com/php-webdriver/php-webdriver/blob/main/lib/JavaScriptExecutor.php))
>    
>    * while we cannot use [PHP intersection types](https://php.watch/versions/8.1/intersection-types) yet, we could have nevertheless used it in PHPDoc (and it would be relatively safe thanks to PHPStan)
> 2. Some functionality (for example `action()` and `isW3cCompliant()`) is not covered by an interface
>    
>    * we would need to do expensive runtime checks to ensure that the instantiated driver has those methods (or `assert()` checks to pass PHPStan, but risk calling undefined methods when assertions are disabled).
> 3. The situation is similar with `RemoteWebElement` (concretion) and `WebDriverElement` (interface) - we use `getDomProperty()` which is not in any interface.
> 
> ~I'll try to suggest some changes to the library.~
> 
> Apparently some are already defined, but they're waiting for the next major version for BC break: https://github.com/php-webdriver/php-webdriver/blob/998e499b786805568deaf8cbf06f4044f05d91bf/lib/WebDriverElement.php#L43

If you to move partial mock-related changes into a separate PR (see the previous comment), then this discussion should happen there.

Anyway, what can we do to make it work? Apparently, we can't rely on the correctness of the `php-webdriver/webdriver` provided interfaces and need to use its concrete interface implementations, which have extra public methods that we use.

src/WebdriverClassicDriver.php
```diff
@@ -91,18 +92,26 @@ class WebdriverClassicDriver extends CoreDriver
...
      * @param string $browserName One of 'edge', 'firefox', 'chrome' or any one of {@see WebDriverBrowserType} constants.
      * @param TCapabilities $desiredCapabilities
+     * @param TDriverInstantiator $driverInstantiator
      */
     public function __construct(
         string $browserName = self::DEFAULT_BROWSER,
         array $desiredCapabilities = [],
-        string $webDriverHost = 'http://localhost:4444/wd/hub'
+        string $webDriverHost = 'http://localhost:4444/wd/hub',
+        ?callable $driverInstantiator = null
```
aik099, line 109: We're in a driver class and the parameter is called `$driverInstantiator`. Shouldn't it be called `$webDriverInstantiator`?

src/WebdriverClassicDriver.php
```diff
@@ -1096,14 +1104,11 @@ private function withWindow(?string $name, callable $callback): void
      */
     private function findElement(
         #[Language('XPath')]
-        string $xpath,
-        ?RemoteWebElement $parent = null
```
aik099, line 1076: Why was the `$parent` parameter (and the code, that uses it in this method) removed?

phpstan.dist.neon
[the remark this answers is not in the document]
aik099: OK.

aik099 (changes requested):
+ see my other responses in other comments:

* https://github.com/minkphp/webdriver-classic-driver/pull/55#discussion_r1902711920
* https://github.com/minkphp/webdriver-classic-driver/pull/55#discussion_r1902716925
* https://github.com/minkphp/webdriver-classic-driver/pull/55#issuecomment-2570566666

.

tests/bootstrap.php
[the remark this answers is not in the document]
uuf6429: I agree that the documentation is a bit vague, but I think it should be safe because:
1. empty string is not considered an invalid url, but rather a blank one: https://3v4l.org/uNvJM (and I doubt they would change that behaviour)
2. in any case, we handle `null`, `false` and `''` in the same way with `?:`
3. that same logic is also there to handle invalid values so if, for example, they change the behaviour to throw an exception, we would need to catch it for invalid input too, not just the default value
4. this is only used for test bootstrap - we have a lot of flexibility (and less responsibility) there

uuf6429:
> Anyway, what can we do to make it work? Apparently, we can't rely on the correctness of the php-webdriver/webdriver provided interfaces and need to use its concrete interface implementations, which have extra public methods that we use.

I've made a an attempt in a different branch, but I didn't push it yet since it depends on this one and it would be too complicated to review. Since this is getting a bit complex, I suggest we organise it so:
1. a PR for refactoring/improving webdriver creation (new PR based on your "unrelated changes" comment)
2. a PR for the phpstan upgrade (this current PR)
3. a PR to attempt switching to depending more on interfaces instead of concretions

src/WebdriverClassicDriver.php
uuf6429: We never used that functionality and since it's a private method and is not worth the extra condition and code to handle it at this point, I decided to remove it.

Removing that extra argument also helps a bit with switching to interfaces.

tests/bootstrap.php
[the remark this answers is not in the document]
aik099: OK.

src/WebdriverClassicDriver.php
aik099: OK.

aik099 (approved):

PR #56, opened 2025-01-04
Implement driver initializer

uuf6429:

PR #57, opened 2025-01-04
Fix various phpstan warnings

uuf6429:
Some warnings will show up if we remove `checkMissingIterableValueType` and `treatPhpDocTypesAsCertain` settings (which makes PHPStan somewhat stricter). 

- [x] ⚠️ Must be rebased on main after #56 is merged.

### DIFF
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -788,11 +788,6 @@ class WebdriverClassicDriver extends CoreDriver
         throw new DriverException('Base driver has not been created');
     }
 
-    protected function getDesiredCapabilities(): array
-    {
-        return $this->desiredCapabilities->toArray();
-    }
-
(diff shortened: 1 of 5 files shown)


src/WebdriverClassicDriver.php
```diff
@@ -340,12 +347,16 @@ public function getAttribute(
...
+     * {@inheritdoc}
+     * @return TElementValue
+     */
     public function getValue(
         #[Language('XPath')]
         string $xpath
     ) {
         $element = $this->findElement($xpath);
-        $widgetType = strtolower($element->getTagName() ?? '');
+        $widgetType = $element->getTagName();
```
uuf6429, line 359: I found out that `getTagName` is guaranteed to return a lowercase string, so the code here is quite redundant (and removing it might additionally fix a bug with multibyte tag names).

PR #59, opened 2025-01-04
Use interfaces as much as possible; switch to actions convenience class

uuf6429:
- [x] ⚠️ Must be rebased on main after https://github.com/minkphp/webdriver-classic-driver/pull/55 is merged.

### DIFF
--- a/src/WebdriverClassicDriver.php
+++ b/src/WebdriverClassicDriver.php
@@ -84,7 +84,7 @@ class WebdriverClassicDriver extends CoreDriver
     private const W3C_WINDOW_HANDLE_PREFIX = 'w3cwh:';
 
     /**
-     * @var TWebDriver|null
+     * @phpstan-var TWebDriver|null
      */
     private ?WebDriver $webDriver = null;
 
@@ -766,7 +766,7 @@ class WebdriverClassicDriver extends CoreDriver
      */
     public function getWebDriverSessionId(): ?string
     {
-        return $this->isStarted() && method_exists($this->getWebDriver(), 'getSessionId')
+        return $this->isStarted() && method_exists($this->getWebDriver(), 'getSessionID')
             ? $this->getAsString($this->getWebDriver()->getSessionID(), 'Session ID')
             : null;
     }
@@ -804,7 +804,7 @@ class WebdriverClassicDriver extends CoreDriver
     }
 
     /**
-     * @return TWebDriver
+     * @phpstan-return TWebDriver
      * @throws DriverException
      */
     protected function getWebDriver(): WebDriver
@@ -988,7 +988,7 @@ class WebdriverClassicDriver extends CoreDriver
      * $this->executeJsOnElement($element, 'return argument[0].childNodes.length');
      * ```
      *
-     * @param TWebDriverElement $element
+     * @phpstan-param TWebDriverElement $element
      * @return mixed
      * @throws DriverException
      */
@@ -1067,7 +1067,7 @@ class WebdriverClassicDriver extends CoreDriver
     }
 
     /**
-     * @param TWebDriverElement $element
+     * @phpstan-param TWebDriverElement $element
      * @throws DriverException
      */
     private function clickOnElement($element): void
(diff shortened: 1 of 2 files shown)


src/WebdriverClassicDriver.php
```diff
@@ -1040,37 +1066,14 @@ private function getWindowHandleFromName(string $name): string
...
-    }
-
-    /**
-     * @throws DriverException
-     */
-    private function mouseOverElement(RemoteWebElement $element): void
-    {
-        $element->getLocationOnScreenOnceScrolledIntoView();
-        $this->getWebDriver()->getMouse()->mouseMove($element->getCoordinates());
+        $element->click();
```
uuf6429, line 1076: I wanted to be consistent with the other actions by using this:
```php
$this->actions()->click($element)->perform();
```
but for some strange reason, it does not work well in firefox, failing many tests:
https://github.com/minkphp/webdriver-classic-driver/actions/runs/12611297356/job/35146807053

I thought it might be because of [the geckdriver-specific fix](https://github.com/php-webdriver/php-webdriver/blob/998e499b786805568deaf8cbf06f4044f05d91bf/lib/Remote/RemoteWebElement.php#L83) in `RemoteWebElement->click()`, but it does not seem to be the case.

My last remaining guess is that `DriverCommand::CLICK_ELEMENT` also waits for the page to finish loading, whereas `DriverCommand::ACTIONS`(W3C)/`DriverCommand::CLICK`(JWP) do not wait.

It may be useful to figure out exactly why, it might be the reason behind various fragile tests we have.

aik099: You don't need to guess. Just do real-time debugging inside a WebDriver code (if you can reproduce a problem locally).

uuf6429: This happens within selenium (and possibly the webbrowser). `DriverCommand::CLICK_ELEMENT` and `DriverCommand::ACTIONS`(W3C)/`DriverCommand::CLICK`(JWP) produce a different request to selenium. Of course I'll dig deeper..

src/WebdriverClassicDriver.php
```diff
@@ -77,7 +83,10 @@ class WebdriverClassicDriver extends CoreDriver
 
     private const W3C_WINDOW_HANDLE_PREFIX = 'w3cwh:';
 
-    private ?RemoteWebDriver $webDriver = null;
+    /**
+     * @var TWebDriver|null
```
stof, line 87: anything using phpstan type aliases should use a `phpstan-` prefix for `@var`, `@param` and `@return` to avoid issues with tools reading phpdoc without knowing about phpstan type aliases, which would treat this `TWebDriver` as a class name

stof (commented):
Can you rebase this PR ?

src/WebdriverClassicDriver.php
```diff
@@ -751,8 +766,8 @@ public function getBrowserName(): string
      */
     public function getWebDriverSessionId(): ?string
     {
-        return $this->isStarted()
-            ? $this->getWebDriver()->getSessionID()
+        return $this->isStarted() && method_exists($this->getWebDriver(), 'getSessionId')
+            ? $this->getAsString($this->getWebDriver()->getSessionID(), 'Session ID')
```
aik099, line 770: You're checking, that the `getSessionId` method exists, but calling `getSessionID`. Please use the same case (based on method declaration).

src/WebdriverClassicDriver.php
```diff
@@ -18,14 +18,18 @@
 use Facebook\WebDriver\Exception\TimeoutException;
 use Facebook\WebDriver\Exception\UnsupportedOperationException;
 use Facebook\WebDriver\Exception\WebDriverException;
+use Facebook\WebDriver\Interactions\WebDriverActions;
+use Facebook\WebDriver\Internal\WebDriverLocatable;
```
aik099, line 22: The `Internal` namespace suggests, that this class/interface shouldn't be used outside of the library and therefore not included in BC promise. But maybe I'm mistaken.

stof:
Looking into this again, I think we should not do that. Having to use internal interfaces and tons of `method_exists` checks to avoid using the `RemoveWebDriver` type while it is part of the public API of the library looks worse to me.

PR #60, opened 2025-01-06
Use PHPStan-specific annotations when using PHPStan alias types

uuf6429:


aik099 (approved):

aik099:
The PR title is misleading. Phpstan types already work, but the way how they're specified breaks other stuff.

I'd suggest rename to `Use PhpStan-specific PhpDoc annotations for data types, that only PhpStan can understand.` or something similar.

oleg-andreyev:
@uuf6429 I assume it's because of "other" IDE then PhpStorm? because PS does understand this type tags

uuf6429:
> @uuf6429 I assume it's because of "other" IDE then PhpStorm? because PS does understand this type tags

Nah, it's about other tools, for example psalm. @stof made a good point about, so I'm fixing it in this PR.

stof:
Psalm actually knows about phpstan type aliases (if the type used in them is supported by psalm, as it converts it to a psalm type alias). However, some other tools don't support it (for instance, `symfony/type-info`)

stof (approved):

aik099 (approved):

PR #61, opened 2025-01-08
Update docker-compose file

uuf6429:
Mainly this is to remove "version", which [has been obsoleted](https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete).

I've additionally changed the indentation as well.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,22 +1,20 @@
-version: "3"
-
 services:
-  selenium:
-    image: ${SELENIUM_IMAGE:-selenium/standalone-chrome:4}
-    hostname: selenium
-    shm_size: 4g
-    environment:
-      VNC_NO_PASSWORD: 1
-      SCREEN_WIDTH: 1024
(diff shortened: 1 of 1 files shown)


uuf6429:
Seems that these sudden failures relate to the newly-added "switch to missing window" test.

I guess this PR could still be merged, but maybe it makes sense to revert that test, at least temporarily. 😞

aik099:
For me locall both "webdriver-classic-driver" (this driver) and MinkSelenium2Drver passed `switching to missing window` test.

aik099:
That's what I call a flaky test. Some of the window operation-related tests are randomly failing on GitHub Actions. I have no idea why.

aik099 (approved):

uuf6429:
I'm starting to get the feeling there might be some problem with the github runners - I also had weird issues when I was trying to set up the chrome driver tests on github.

I'll investigate that later.

